PR #22, opened 2020-05-20
Implement OMEX archive export via libCombine native bindings

gmarupilla:


danv61:
Hi Akhil, it's hard to say by just looking at the code. What about merging your changes into alpha and then we'll test how the project updates, and then we test the export itself?
In the current alpha a simple exporting to sedx works already, so we'll try the same thing for omex. If all is well, we'll scrap the sedx export and only keep the new omex export.

gmarupilla:
Yes, I can merge, but do I need to merge to the master(main) branch?

PR #90, opened 2021-09-15
Enhanced specifications

jonrkarr:
This PR expands the specifications of VCell. We've done the same for all other simulation tools.
- Model changes
- Simulation types
- Output dimensions
- Output variables

@moraru @danv61  which of the following does VCell support? Please correct the specifications accordingly
- [ ] SED-ML compute model changes
- [ ] SED-ML add/remove/change XML model changes
- [ ] SED-ML set value model changes (used with repeated tasks)
- [ ] Observables for parameter values (i.e., parameters set by assignment rules)
- [ ] Observables for reaction fluxes
- [ ] Observables for compartment sizes

### DIFF
--- a/biosimulators.json
+++ b/biosimulators.json
@@ -179,6 +179,40 @@
           "supportedFeatures": []
         }
       ],
+      "modelChangePatterns": [
+        {
+          "name": "Change component attributes",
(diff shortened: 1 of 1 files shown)


moraru:
@jonrkarr I merged into master and released v.20 (together with some other bug fixes and updates)

PR #109, opened 2021-11-08
fix: fixed schema for BioSimulators logs

jonrkarr:
The schema for logs uses `type` rather than `category`. We previously documented this incorrectly. 

A tool for validating logs is now available here. This is also now incorporated into the BioSimulators test suite.
https://api.biosimulations.org/#/Logs/LogsController_validateLog

@moraru @danv61 please merge and release another version of VCell

### DIFF
--- a/vcell-cli-utils/vcell_cli_utils/status.py
+++ b/vcell-cli-utils/vcell_cli_utils/status.py
@@ -75,10 +75,10 @@ def status_yml(omex_file: str, out_dir: str):
             #outputs_dict["outputs"][report].update({"status": "QUEUED"})
 
         for task in task_list:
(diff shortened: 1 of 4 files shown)


jonrkarr:
@moraru @danv61 I've marked VCell 7.4.0.24 as invalid because the logs fail the newly added validation. This should be fixed by this PR. Other than that, VCell looks to be working in runBioSimulations. All 5 example runs that I ran with VCell this afternoon passed (except for the invalid logs).

VCell will automatically be enabled again once this PR is merged and a new version of VCell is released to BioSimulators.

moraru:
merged and will release/publish new version

Issue #229, opened 2022-07-22
SBML/COMBINE export with roundtrip validation


jcschaff:
The VcmlOmexConverter used in exporting SBML/SEDML (together in an .omex file - the COMBINE archive) did not verify that the resuting .omex file could be re-parsed by VCell.  This round-tripping needs to be put in place to shorten the development cycle and fail an export which is invalid.

jcschaff:
created branch and introduced:

- introduced logging (using log4j2) to replace verbose print statements in SBML/SEDML import/export classes.
- omex file validation by immediately re-reading the exported omex files to perform basic validation (omex file parse doesn't fail).
- fixed SEDML export of "initCount" parameters (for stochastic VCell applications).
- better error message for SBML export with membrane voltage.
- fixed SBML import of 2D/3D geometries with respect to subvolumes and membranes.

jcschaff:
Known Problem:
In SBML import/export related to AssignmentRules variables not being imported or exported properly for diffusion coefficient.  This will be fixed in a follow up issue.

jcschaff:
see  [PR 230](https://github.com/virtualcell/vcell/pull/230)

PR #230, opened 2022-07-22
round-trip export/import for COMBINE archives (.omex files)

jcschaff:
- introduced logging (using log4j2) to replace verbose print statements in SBML/SEDML import/export classes.
- omex file validation by immediately re-reading the exported omex files to perform basic validation (omex file parse doesn't fail).
- fixed SEDML export of "initCount" parameters (for stochastic VCell applications).
- better error message for SBML export with membrane voltage.
- fixed SBML import of 2D/3D geometries with respect to subvolumes and membranes.

see https://github.com/virtualcell/vcell/issues/229

### DIFF
--- a/vcell-cli/pom.xml
+++ b/vcell-cli/pom.xml
@@ -79,6 +79,20 @@
 
     <dependencies>
         <dependency>
+            <groupId>org.apache.logging.log4j</groupId>
+            <artifactId>log4j-api</artifactId>
+            <version>2.17.1</version>
+        </dependency>
+        <dependency>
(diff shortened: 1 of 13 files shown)


jcschaff:
note: the stdout and stderr outputs from the CLI have been largely ported to logging (log4j2).  You may filter to see only warnings and errors by editing, but not committing, [vcell-cli/src/main/resources/log4j2.xml)](https://github.com/virtualcell/vcell/pull/230/commits/c56385f4df3ffe23e4bfb81acc4a310091e7ea9d)

Issue #234, opened 2022-07-26
CLI for biosimulations should be compliant with their spec


jcschaff:
1. CLI for biosimulations should be compliant with their spec (correct flags, with --help, --debug, --quiet)
2. extend the CLI for a dedicated biosimulations command (to assure the interface is correct and stable)
3. CLI when run out of Docker the biosimulations command should be inserted by default
4. help command will still expose the whole CLI suite.

jcschaff:
fixed poetry problem in Dockerfile (and Singularity)

jcschaff:
merged, closed

PR #235, opened 2022-07-26
CLI biosimulations command (is default command from docker)

jcschaff:
[see Issue 234: CLI for biosimulations should be compliant with their spec](https://github.com/virtualcell/vcell/issues/234) for description


jcschaff:
fixed poetry path problem in Dockerfile (and singularity)

CodeByDrescher (approved):

PR #241, opened 2022-07-29
Change in poetry installation via docker

CodeByDrescher:
Changed poetry installation to use apt instead of curl-ing and running a python script


jcschaff (approved):
code looks right - Approved!  @CodeByDrescher please merge if it passes CI.

PR #257, opened 2022-08-08
Overrides to sedml

danv61:


moraru:
@danv need to verify syntax and execution with other simulator for test model containing mix of scanned parameters and dependent functions

vcell-core/src/main/java/org/vcell/sedml/SEDMLExporter.java
```diff
@@ -512,6 +538,7 @@ private void translateBioModelToSedML(String savePath, String sBaseFileName, boo
 
 							for (String unscannedParamName : unscannedParamHash.values()) {
 								SymbolTableEntry ste = getSymbolTableEntryForModelEntity(mathSymbolMapping, unscannedParamName);
+								assert ste != null;
```
jcschaff, line 541: throw new RuntimeException("failed to retrieve vcell parameter from unscanned SEDML parameter "+unscannedParamName+" using mathSymbolMapping);

jcschaff: we should probably avoid assert statements in production code.

moraru: I would clean those up once we also do proper log4j logging in SEDMLexporter... class also needs some refactoring of giant method that does the actual export

jcschaff (approved):
approved, waiting for integration of 3 branches to do final testing.

Issue #239, opened 2022-07-28
VCell CLI needs to re-adhere to Biosimulators' CLI standards.


CodeByDrescher:
### Background
- As a result of recent changes to VCell CLI, the current state does not adhere to the Biosimulators' standards for CLI applications. As this is important to us, we need to fix the issues locally before sending VCell back to the Biosimulators' GitHub repo for retesting.

### Tasks:
1. get the biosimulators_test_suite executable operational on vcell-test-suite vm box.
2. determine if existing fixes since the original discovery are sufficient, or if there are further / new bugs to defeat.

CodeByDrescher:
Converted to a milestone

PR #258, opened 2022-08-08
SBMLSymbolMapping and SBMLImporter refactor

jcschaff:
see #239 

## Goals
- capture mapping of SBML SBase to VCell SymbolTableEntry objects during SBML import for SEDML processing.
- simplify SBMLImporter by separately constructing VCell objects, recording mapping to SBase, recording SBML expressions, translate SBML expressions in context of VCell objects
- for VCML->SBML->VCML, verifying that the re-imported SBML is mathematically equivalent to the original.
- fix issues in SBMLImporter, SBMLExporter and elsewhere by systematic testing (manual and CI unit tests).

### Status
- one complex and old vcell spatial model (actin comet tails) round-trips with equivalent math with no hacks.
- added "fast reactions", "inside/outside" structures, and "volume sampling size" for geometry for lossless round-trips.
- added over 500 unit tests from SBML Test Suite - runs in CI very quickly (import only, doesn't run simulations).

### Known issues
- needs more testing in CI (sample the rest of SBML Test Suite, add SBML Spatial test suite, add some round-trip tests for VCell models).
- external boundary conditions types (e.g. Value/Flux/Mixed) at x,y,z boundaries of geometry for membrane domains in SBML can be chosen arbitrarily.  In VCell, for FiniteVolume solvers (semi-implicit and full implicit) are tied to one of the neighboring volumes in unexpected ways (a historical artifact).  We need to warn/error for SBML models that we cannot support.
- SBMLImport of math with event triggers is incomplete (see SBML Test Suite test cases 436 and 437).


jcschaff:
This PR is very large - sorry.  Most interesting stuff is in SBMLImporter - which is smaller and simpler - but still needs a pass to clarify design with some renaming.

Reviewers Ion and Dan are not expected to read every line - but questions or comments are welcome.

jcschaff:
merging at the suggestion of Ion ... and realistically, a proper review is not possible with the number of changes in this PR ... smaller next time.

PR #263, opened 2022-08-11
added backup library loader to OmexHandler and NativeLib

CodeByDrescher:
### Background
When setting up my new workstation, I came across the same native library bug I had encountered on my old system. After discovering a work around with the help of Jim, I created a simple "backup" loader in case @gmarupilla's code fails. 

### VERIFICATION NEEDED
This hasn't been tested on all platforms. It currently has only been tested on Linux (Ubuntu 20); Before this branch is accepted, we need confirmation that it works on:
- Windows (10)
- Mac

Once we have confirmation, we can pull this in.

When we have some time to tackle tech debt, this whole system may be moved to its own class that is called near time of execution (or at least somewhere more appropriate than buried where it is).

### DIFF
--- a/vcell-cli/src/main/java/org/vcell/cli/run/OmexHandler.java
+++ b/vcell-cli/src/main/java/org/vcell/cli/run/OmexHandler.java
@@ -2,6 +2,9 @@ package org.vcell.cli.run;
 
 import cbit.vcell.resource.NativeLib;
 import cbit.vcell.resource.ResourceUtil;
+
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 import org.sbml.libcombine.CaContent;
 import org.sbml.libcombine.CaListOfContents;
 import org.sbml.libcombine.CaOmexManifest;
@@ -23,6 +26,8 @@ public class OmexHandler {
     String outDirPath;
     CombineArchive archive;
 
+    private final static Logger logger = LogManager.getLogger(OmexHandler.class);
+    
     // Assuming omexPath will always be absolute path
     // NB: Need to convert class to use Log4j2
     public OmexHandler(String omexPath, String outDir) throws IOException {
@@ -31,14 +36,14 @@ public class OmexHandler {
                 ResourceUtil.setNativeLibraryDirectory();
                 NativeLib.combinej.load();
             } catch (Exception e){
-                System.err.println("Unable to link to native 'libCombine' lib, check native lib. Attemping alternate solution...");
+            	logger.error("Unable to link to native 'libCombine' lib, check native lib. Attemping alternate solution...");
                 NativeLib.combinej.directLoad();
             }
         } catch (UnsatisfiedLinkError ex) {
-            System.err.println("Unable to link to native 'libCombine' lib, check native lib: " + ex.getMessage());
+            logger.error("Unable to link to native 'libCombine' lib, check native lib: " + ex.getMessage());
             throw ex;
-        } catch (Exception ex) {
-            System.err.println("Error occurred while importing libCombine: " + ex.getMessage());
+        } catch (RuntimeException ex) {
+            logger.error("Error occurred while importing libCombine: " + ex.getMessage());
             throw ex;
         }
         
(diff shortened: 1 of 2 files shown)


CodeByDrescher:
Confirmed branch works on Windows; Only need Mac confirmation

vcell-cli/src/main/java/org/vcell/cli/run/OmexHandler.java
```diff
@@ -24,18 +24,24 @@ public class OmexHandler {
...
         } catch (UnsatisfiedLinkError ex) {
             System.err.println("Unable to link to native 'libCombine' lib, check native lib: " + ex.getMessage());
-            System.exit(1);
-        } catch (Exception e) {
-            System.err.println("Error occurred while importing libCombine: " + e.getMessage());
-            System.exit(1);
+            throw ex;
+        } catch (Exception ex) {
+            System.err.println("Error occurred while importing libCombine: " + ex.getMessage());
+            throw ex;
```
jcschaff, line 42: looks like OmexHandler constructor declares throwing IOException, but this throws Exception as well (how does this compile?)

vcell-core/src/main/java/cbit/vcell/resource/NativeLib.java
```diff
@@ -69,4 +75,37 @@ public boolean isDone( ) {
...
+		return fileName + "." + this.getAppropraiteLibrarySuffix();
+	}
+
+	private String getAppropraiteLibrarySuffix(){
+		String suff;
+		OperatingSystemInfo osi = OperatingSystemInfo.getInstance();
+		if (osi.isWindows()){
+			suff = "dll";
+		} else if (osi.isMac()){
+			suff = "jnilib";
```
jcschaff, line 102: dylib is the modern suffix for mac ... we don't have any because our libraries are old.   Maybe it makes sense to change this to 'getAppropriateLibrarySuffixes()' and return new String[] { "jnilib", "dylib" }

vcell-core/src/main/java/cbit/vcell/resource/NativeLib.java
```diff
@@ -69,4 +75,37 @@ public boolean isDone( ) {
...
+		String installDir = ResourceUtil.getVCellInstall().getAbsolutePath();
+		String nativeLibDir = "nativelibs";
+		String osDir = osi.getNativeLibDirectory();
+		String libFileName = this.getLibraryFileName();
+
+		Path path = Paths.get(installDir, nativeLibDir, osDir, libFileName);
+		return path.toAbsolutePath().toString();
+	}
+
+	private String getLibraryFileName(){
```
jcschaff, line 89: public String[] getLibraryFileNames() {
....
ArrayList<String> filenames = new ArrayList<>();
for (String suffix : getAppropriateLibrarySuffixes()){
filenames.add(filename + "." + suffix);
}
return filenames;

vcell-core/src/main/java/cbit/vcell/resource/NativeLib.java
```diff
@@ -69,4 +75,37 @@ public boolean isDone( ) {
...
 
+	private String generatePath(){
+		OperatingSystemInfo osi = OperatingSystemInfo.getInstance();
+		String installDir = ResourceUtil.getVCellInstall().getAbsolutePath();
+		String nativeLibDir = "nativelibs";
+		String osDir = osi.getNativeLibDirectory();
+		String libFileName = this.getLibraryFileName();
+
+		Path path = Paths.get(installDir, nativeLibDir, osDir, libFileName);
+		return path.toAbsolutePath().toString();
```
jcschaff, line 86: before return, loop through possible filepaths to see if any exist (e.g. File.exists()) before returning the path.

jcschaff (changes requested):
Nice job.  A few small comments.  logging vs printing to stderr and changes to support alternative library suffixes on macos.

vcell-cli/src/main/java/org/vcell/cli/run/OmexHandler.java
CodeByDrescher: I'm 95% sure its because the caught exceptions that can make it there are all IOException; So even if we use the more generic Exception, java knows its safe.

I'll change line 40 to be explicitly IOException anyway; cleaner choice.

CodeByDrescher: Ahh, I was wrong; There's code lower down that explicitly need IOException; any exception caught by the line 40 Exception statement will be RuntimeExceptions; I fixed that line accordingly

jcschaff (approved):
Thanks for putting this together, I tested it on macos.

PR #268, opened 2022-08-15
Restored deprecated single archive execution and added cleaning function

CodeByDrescher:
Biosimulator's cli standard requires the ability to execute a single archive rather than a directory of n archives (n is a natural number). This feature had been recently deprecated accidentally, but will now be supported. Additionally, proper cleaning of output directory is performed.

### DIFF
--- a/vcell-cli/src/main/java/org/vcell/cli/CLIUtils.java
+++ b/vcell-cli/src/main/java/org/vcell/cli/CLIUtils.java
@@ -39,6 +39,17 @@ public class CLIUtils {
         }
     }
 
+    public static void cleanRootDir(File outdir){
+        // If this could be done without hard coding it'd be preferable.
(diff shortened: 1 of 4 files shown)


vcell-cli/src/main/java/org/vcell/cli/CLIUtils.java
```diff
@@ -39,6 +39,17 @@ private static void deleteRecursively(File f) throws IOException {
         }
     }
 
+    public static void cleanRootDir(File outdir){
+        // If this could be done without hard coding it'd be preferable.
```
jcschaff, line 43: do you only want to remove these files or all files in the directory?  This is a somewhat unexpected side-effect.  It is one thing to overwrite without asking, it is another to remove files.

jcschaff: as we discussed it may be good to remove only what is appended.

jcschaff (approved):
looks good

PR #272, opened 2022-08-17
Added custom Exception VCLoggerException and refactored accordingly

CodeByDrescher:
This will need more work in the future; a lot of cleanup is needed to fix the prevalent misuse of exceptions.

### DIFF
--- a/vcell-cli/src/main/java/org/vcell/cli/sbml/MemoryLogger.java
+++ b/vcell-cli/src/main/java/org/vcell/cli/sbml/MemoryLogger.java
@@ -32,6 +32,7 @@ public class MemoryLogger extends VCLogger {
             lowPriority.add(msg);
         }
         System.err.println(p + " " + et + ": "  + message);
+        throw new VCLoggerException(message);
     }
 
 }
--- a/vcell-cli/src/main/java/org/vcell/cli/sbml/MemoryLogger.java
+++ b/vcell-cli/src/main/java/org/vcell/cli/sbml/MemoryLogger.java
@@ -32,7 +32,6 @@ public class MemoryLogger extends VCLogger {
             lowPriority.add(msg);
         }
         System.err.println(p + " " + et + ": "  + message);
-        throw new VCLoggerException(message);
     }
 
 }
--- a/vcell-cli/src/main/java/org/vcell/cli/sbml/MemoryLogger.java
+++ b/vcell-cli/src/main/java/org/vcell/cli/sbml/MemoryLogger.java
@@ -1,6 +1,7 @@
 package org.vcell.cli.sbml;
 
 import cbit.util.xml.VCLogger;
+import cbit.util.xml.VCLoggerException;
 
 import java.util.ArrayList;
 import java.util.List;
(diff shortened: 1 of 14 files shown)


vcell-client/src/main/java/cbit/vcell/client/BNGWindowManager.java
```diff
@@ -148,7 +149,7 @@ public void importSbml(String bngSbmlStr) {
 
 			@Override
 			public void sendMessage(Priority p, ErrorType et, String message)
-					throws Exception {
+					throws VCLoggerException {
 				System.err.println("LOGGER: msgLevel="+p+", msgType="+et+", "+message);
 				if (p == VCLogger.Priority.HighPriority) {
 					throw new RuntimeException("Import failed : " + message);
```
jcschaff, line 155: this is the whole point of the Exception.

if (p == VCLogger.Priority.HighPriority) {
   throw new VCLoggerException(...);
}

if we are going to only use runtime exceptions ... (if that is the only thing being thrown right now within sendMessage() then we can remove the throws clause completely.

jcschaff (changes requested):

vcell-cli/src/main/java/org/vcell/cli/sbml/MemoryLogger.java
```diff
@@ -31,6 +32,7 @@ public void sendMessage(Priority p, ErrorType et, String message)
             lowPriority.add(msg);
         }
         System.err.println(p + " " + et + ": "  + message);
+        throw new VCLoggerException(message);
```
jcschaff, line 35: Here I specifically wanted to only log problems and keep going (without throwing exception).

jcschaff: .. and definitely wouldn't want to throw an exception for medium or low priority messages.

vcell-client/src/main/java/cbit/vcell/client/TranslationLogger.java
```diff
@@ -93,10 +94,10 @@ public void sendAllMessages() {
 
 	@Override
 	public void sendMessage(Priority p, ErrorType et, String message)
-			throws Exception {
+			throws VCLoggerException {
 
 		if  (p == null ||et == null ) {
-			throw new IllegalArgumentException("Invalid params for sending translation message.");
+			throw new VCLoggerException(new IllegalArgumentException("Invalid params for sending translation message."));
```
jcschaff, line 100: Comment:  you don't have to wrap subclasses of RuntimeException in VCLoggerException if it doesn't add clarity of intent or control flow.  Recall that RuntimeExceptions never have to be declared.  Your call ...

vcell-core/src/test/java/org/vcell/sbml/TLogger.java
```diff
@@ -31,6 +32,7 @@ public void sendMessage(Priority p, ErrorType et, String message)
 			lowPriority.add(msg);
 		}
 		System.err.println(p + " " + et + ": "  + message);
+		throw new VCLoggerException(message);
```
jcschaff, line 35: again here the intent is to collect messages for later processing, but not to throw an exception and abort.

jcschaff (changes requested):

jcschaff (approved):
looks good, thanks for taking the initiative to make things better.

PR #276, opened 2022-08-18
Sedml importer

moraru:
XMLHelper giant single method for SED-ML import was delegated to dedicated SEDMLImporter class.

- Import logic was reorganized and broken down into orderly sequence.
- New major functionality:
 -- support RepeatedTask (previously ignored)
 -- check Model changes to see if they can be imported as overrides into another imported BioModel
 -- support Functions in model changes (previously flattened, now preserved) and repeated tasks (previously ignored)
- Various small bug fixes on both import and export throughout

All VCell Simulation math override functionality (values, scans, functions, functions of scans, in any combination) now fully supported and everything round-trip as such in single biomodel/simcontext.

Limited testing, based on shared BioModel "--0-export-scan-test" and variations thereof.


moraru:
Notes and updates:
- The only conflict is trivial, a straight method swap in XMLHelper class
-  I've done more extensive testing and here are the expectations:
  -- it should work for all deterministic apps
  -- it should also work for all stochastic apps unless there are overrides involving initial concentrations (math symbol mapping is at fault in this case)
  -- it cannot yet be properly tested for spatial apps because even the most simple models fail on export if there are overrides (sedml exporter and sbml exporter are at fault in this case)

Since none of the above limitations are related to the actual code of this PR, I think it should be good to go to merge into master.

vcell-admin/resources/log4j2.xml
```diff
@@ -0,0 +1,31 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<Configuration status="WARN" name="XMLConfigTest" packages="org.vcell.*">
+    <Properties>
```
jcschaff, line 3: @CodeByDrescher take a look at this - FYI this will show up in the class loader as a default (not sure if JSBML or vcell-admin version will be picked up as the default.

Maybe we should put a default into vcell-core instead.

I'll accept this today - but we should rethink it.

jcschaff (approved):
looks good to me.  we should revisit the log4j2.xml strategy - not a blocker.
Approved.

vcell-admin/resources/log4j2.xml
moraru: @jcschaff @CodeByDrescher sorry, that was for my own running config, I accidentally added to Git staging

moraru: I'll remove before merge

PR #278, opened 2022-08-19
Cli elipses bug fix

CodeByDrescher:


CodeByDrescher:
This should fix the last few bugs in bsts, but that's what I thought the past few times.

PR #281, opened 2022-08-22
Sbml exporter validate

jcschaff:
Refactor round-trip validation testing to isolate VCell->SBML->VCell round tripping.
During testing round some problems with MathOverrides and Image-based Geometry.

- fixed image-based geometry round tripping (when VCPixelClass names clash ... they are a don't-care).

- automatically replace MathOverrides constant names (which are mangled with units) when they are equivalent (e.g. Constant names 'abc_init_uM' and 'abc_init_umol_l_1').

- automatically replaces transformed MathOverrides due to unit system transformation (e.g. from VCell->SBML->VCell).

- added unit test ModelUnitConverterTest to illustrate transformation and make sure we recognize if it breaks in the future

- known problem:  The unit transformed math overrides don't cancel KMOLE as needed (KMOLE is simplified away before it can be cancelled).  So the round-trip override values are ugly but correct.


jcschaff:
Sorry so large, should work well.

PR #292, opened 2022-08-23
Changed comparison of full child URI to ending of child URI

CodeByDrescher:
Fix to allow VCell CLI to properly process nested sed-ml docs.

Thanks to everyone for the help.


CodeByDrescher:
Solution needs to account for same model name, but different subdirs

PR #299, opened 2022-08-25
validate=true for SBML export from VCell UI.

jcschaff:
one-liner needed for production VCell.  round-trip validation of SBML export from VCell UI.


jcschaff:
no review needed, already discussed as a group.  Should always validate SBML export (including in UI).

PR #300, opened 2022-08-25
SBMLExporter validate compares transformed SimContext (e.g. stochastic)

jcschaff:
SBMLExporter performs round-trip testing to make sure re-imported SBML models have and equivalent mathematical representation.  SBMLImporter chooses between non-spatial deterministic and spatial deterministic.  But while exporting stochastic applications (SimulationContexts), we must transform the SimulationContext to also be stochastic or the maths will be of different type.

This is now performed in SBMLExporter.roundTripValidation() - so we can validate the exporting and reimporting of stochastic applications.


jcschaff:
did local build/test, looks good.

PR #301, opened 2022-08-26
Sedml roundtrip

moraru:
This PR fixes all known issues for round-tripping MathOverrides in simulations from NETWORK_DETERMINISTIC applications (non-spatial and spatial):

- primary changes to SEDML import and export
- includes cherry-picked commits from @jcschaff from branch map_symbol_unitsize

It also temporarily disables the attempts to import model changes as simulation overrides for stochastic applications, which present significant problems. For stochastic applications changes are being applied to the SBML prior to import.  This has a current limitation that compute changes (which become initial assignments) are lost.


jcschaff (approved):
looks good, approved.

PR #305, opened 2022-08-29
Make Analytic Geometry exp compare tolerate of pow(x,2) vs x^2

jcschaff:
analytic geometry expression comparison in round-trip testing would fail to recognize the equivalence of pow(x,2) and x^2.  Howerver, during VCell->SBML->VCell round-trip, the MathML processing would not retain the original form and VCell would declare that geometries (and hence the mathDescriptions) were not equivalent.

AnalyticGeometry.exp are now compared using ExpressionUtils.functionallyEquivalent() instead of Expression.compareEquals() which is very strict on the exact syntax.


jcschaff:
This fixed round tripping of at least 10 published models.

PR #308, opened 2022-08-29
fix NPE when unmapped compartments exported as SBML

jcschaff:
exporting Spatial SBML from VCell failed with a NullPointerException when there were unmapped compartments (eg. Features and/or Membranes which were not mapped to geometric domains.

Changes were made to tolerate incompletely mapped vcell structures (e.g. compartments).


jcschaff:
this change fixed the SEDML export of 6 published VCell models.

jcschaff:
test

PR #310, opened 2022-08-31
Stoch count conc overrides

jcschaff:
Ion's test model with stochastic application and math overrides failed to transform initial conditions (or corresponding math overrides) between concentration and count.  This is needed for SBML/SEDML, but also just for Virtual Cell.

Added unit tests for Ion's model for concentration->count and for count->concentration transformations.

In order to simplify, ExpressionUtils.simplifyUsingJSCL() employs the basic computer algebra system JSCL which is bundled within VCell to cancel terms.  This new capability has some unit tests as well.


jcschaff:
builds/tests locally.

Issue #316, opened 2022-09-01
JSCL simplify expression slow for large powers


jcschaff:
in PR #314 ExpressionUtils.simplifyUsingJSCL() was improved by passing integer powers as integers (e.g. a^2.0 became a^2) - which allowed more intelligent expand().simplify().factor() within the JCSL computer algebra library.  However, for large powers, expanding these expressions becomes very expensive.

For example: the following expression takes a very long time to simplify (over a minute, didn't wait for completion.

Expression exp = new Expression("(((V32 / ((k32 ^ 3.0) + (Ca_nM ^ 3.0))) + b32) * ((Vm32 * (Ca_nM ^ 7.0) / ((km32 ^ 7.0) + (Ca_nM ^ 7.0))) + bm32))")
Expression simplifiedExp = ExpressionUtils.simplifyUsingJSCL(exp);

A reasonable approach would be to only pass small powers to JSCL as integers, and have large powers passed as floating point (treated differently).

jcschaff:
two strategies are employed, (1) limit length of expression (including intermediate expressions within JSCL), and (2) limit ability of JSCL to expand exponents as polynomials to powers <= 2.

logger.warn() will log if any individual simplification takes more than 1.0 second.

PR #319, opened 2022-09-01
limit expression size for JSCL simplify, fail to exp.flatten(), log t>1s

jcschaff:
see Issue #316


jcschaff:
built/passed tests, greatly improved performance.

PR #360, opened 2022-09-14
Dan repeated tasks3

danv61:
Creating the .csv files, the hdf5 file and plots for parameter scans (repeated tasks) using the existing python code (adapted) 
This is a temporary solution to allow batch testing for published / public models. However, the datasets are flat.


moraru:
This will need more work to create the n-dimensional HDF5 data as current data output is flat (2D sequential).
Current work though good to bring into master.

PR #427, opened 2022-09-29
Ion import/export fixes

moraru:
Large PR that:
- implements batch import functionality
- fixes a plethora of issues discovered during batch import testing (these are a mix of export and import bugs)

Notes:
- a few fixes are temporary workarounds
- logging is minimal and needs to be expanded

I will add more details on the individual issues and fixes, but for now the summary is:
- 100% of exported omex files (total 580) are successfully imported with this codebase
- 563 out of 580 come back as a single vcml/biomodel with all applications/simulations/overrides
-  the remaining 17 that generate more than one biomodel have the same cause: math overrides of an overall reaction kinetic rate (J/LumpedJ) when it is numeric/constant;; this is actually incorrectly exported and will need fixing on both export side (should not be changeAttribute but changeXML) and import side (should check for potential override-compatible changeXML); I will create a separate issue for this


moraru:
Note: this also includes fixes to jMathML which has been brought in as source - needs Maven update projects due to pom.xml changes.

PR #445, opened 2022-10-08
Added temp fieldSimulationVersion to Simulation constructors

CodeByDrescher:
The nullPointerException being caused in recent versions comes down to a change from using TempSimulation to Simulation, and failing to assign a version; On linux systems, a value from this null field would be queried, and subsequently vcell would crash.


jcschaff:
reopening this PR - The commit was reverted - had undesirable side effects (see #816)

Issue #538, opened 2022-11-02
Assignment rules not handled properly during round trip


moraru:
Export/import of explicit assignment rules for species and global parameters are being skipped. Depending on situation, this may be caught by math equivalency testing or by math overrides checking, or not at all.

Desired behavior:
- on export, make sure that explicit assignment rules present in an application are being parsed and not skipped for all targetable variables (global params, species "init", comp size), and take precedence over default values
- on import, we do not have to reconstitute them as explicit assignment rules (this no need to add VCell-specific annotations during the export phase), but just make sure they end up with the right expression in the proper place (global param definition, species init, comp size)

moraru:
@jcschaff check with @danv61 who discovered this issue and is currently making changes to the same area of code while working to fix the duplicate id issue (#519)

danv61:
Export is fixed now in e603d68

danv61:
Fixed in #617 (commit ce518ed)

Issue #519, opened 2022-10-31
roundtrip fails due to duplicate id


moraru:
This affects 6 simcontexts in public models.
Example: biomodel_72815317
May be related to #450

danv61:
Caused by mangling of a name like "a-b" when some other entity is already named "a_b" (happens in SBMLExporter)

danv61:
Fixed in commit e2b4b82

PR #541, opened 2022-11-02
Dan duplicate id bugs

danv61:
Duplicated sid caused by mangling of entities with a name pattern like "a-b" when another entity is named "a_b".


danv61:
I roundtripped with exporting to omex and importing back to vcml, everything was successful. I'm merging this into master, it's definitely an improvement. 
It's addressing issues:
roundtrip fails due to duplicate id #519
Assignment rules not handled properly during round trip #538

danv61:
git commit e603d68

Issue #593, opened 2022-11-21
Relative ratio size calculations can lead to invalid math


moraru:
Compartment volume and surface calculations for legacy models that use relative sizes instead of absolute sizes were changed to fix bugs related to overriding size ratio parameters (see #357 and related issues and PRs).

This leads to occasional circular references in generated math which is invalid (and as a result all simulations as well as all exports fail).

Example: biomodel_17028306
Function  Compartment::Size_Cell	 ( - Size_Cell * VolFract_PM / ( - 1.0 + VolFract_PM));

moraru:
Fix does not seem to work - circular reference still occurring after regenerating math in two published models:
biomodel_17028306
biomodel_14647285

PR #597, opened 2022-11-22
fix other place where structure parameters are not bound to simContext

jcschaff:
#593


jcschaff:
in .refreshDependencies(), StructureMappingParameter expressions were also incorrectly being bound to the wrong SymbolTable, StructureMapping instead of SimulationContext

Issue #679, opened 2022-12-14
fix CI deploy with opt and batch singularity images


jcschaff:
fix CI deploy with opt and batch singularity images

final change I think.

PR #680, opened 2022-12-14
final deploy step needs to work with both singularity images.

jcschaff:
see #679


jcschaff:
final deploy CI change only, not waiting for passing tests.

Issue #695, opened 2022-12-16
Refactoring choosing between export to sedml as vcml vs sbml.


danv61:
Using overriden SedmlExtensionFilter.requiresMoreChoices() and askUser() invoked in ChooseFile

danv61:
Fixed in commit 90ed845, PR #696

PR #696, opened 2022-12-16
Refactoring choosing between export to sedml as vcml vs sbml.

danv61:
Using overriden SedmlExtensionFilter.requiresMoreChoices() and askUser() invoked in ChooseFile


danv61:
Fixed issue #695

Issue #761, opened 2023-01-18
Parameter estimation feedback large reports and stopping


jcschaff:
Parameter estimation feedback improvements:
1. as COPASI's parameter estimation reports get very large (a row is added each time the objective function is reduced), feedback performance degrades and can ultimately timeout.
   a. need to more efficiently parse this file (e.g. subsample, or 'tail' file)
   b. need to reduce size of corresponding message which is communicated back to user.
2. if user prematurely stops parameter estimation, the best parameter values are not displayed to user.
3. ideally, a semi-log graph of objective function vs iteration would be helpful (but may be out of scope for this issue).

PR #762, opened 2023-01-18
761 improved parest feedback

jcschaff:
see #761


jcschaff:
Remaining issue:
particle swarm size is not implemented in optUtils.py (`CopasiOptimizationParameterParamType.SWARM_SIZE`)

Issue #333, opened 2022-09-06
renamed reaction parameters revert to predefined names on save


moraru:
Any built-in kinetic parameter that is being renamed looses the new name upon save.

PR #776, opened 2023-02-03
Fixed renamed reaction parameters revert to predefined names on save.

danv61:
Addressing issue #333


vcell-core/src/main/java/cbit/vcell/model/Kinetics.java
```diff
@@ -1129,14 +1132,14 @@ private ArrayList<KineticsParameter> getKineticsParametersFromTokens(String kine
 		}
```
jcschaff, line 1133: All existing formats may not be documented with comments here, we have to tread carefully when we change the parsing logic.

The setName() code seems harmless ... but consider that this functionality used to work - so unless someone previously changed the parsing code - the parsing code is not the problem.

jcschaff (commented):
left a few comments, will look more closely tonight.

jcschaff:
@danv61 fixed the problem, I reverted the parsing change and some additional helpful comments - then merged/closed this PR.

Issue #798, opened 2023-02-17
BMDB import - adding 1 to species names


vcellmike:
For some models BMDB import adds 1 to all species names. However, downloading SBML file from the database and importing it to VCell through File->Import preserves the correct names. 

In VCell search for Padala2017- ERK, PI3K/Akt and Wnt signalling network (PTEN mutation). Import - all species will get 1 at the end. However, check the same model https://www.ebi.ac.uk/biomodels/BIOMD0000000655 . Save SBML file as .xml, and import to VCell as File => Import. After import, all species have proper names without 1.

danv61:
Commit cb0ac5c

PR #810, opened 2023-02-22
BMDB import adding 1 at the end of some names.

danv61:
When trying to use the SBML name instead the SBML id as VCell name


danv61:
Issue #798

Issue #866, opened 2023-04-20
SedML Validator fails when encounters empty listOf elements.


danv61:

danv61:
We now verify for emptiness all lists: listOfParameters, Variables, Ranges, Changes, SubTasks, DependentTasks, Curves, Surfaces, DataSets.
We don't check for emptiness the lists of Models, Simulations, Tasks, DataGenerators, Outputs. For all these, the validator will be right to complain, I think that nothing will work if any of these is  empty.

danv61:
commit f6b7274

danv61:
PR 867

PR #867, opened 2023-04-20
We don't export empty lists to SedML to humor the validator.

danv61:


danv61:
Issue #866, commit f6b7274

Issue #903, opened 2023-06-09
test biosimulators_vcell against biosimulations projects with known results


jcschaff:
### Summary
VCell's submission of its simulator, biosimulators_vcell must be capable of running those biosimulations.org projects which it claims to support, or at least give a good message that a biosimulations.org end-user can understand.

### Approach
To ensure quality and catch problems early, we need a quantitative unit/integration test for VCell's OMEX execution of Biosimulations Projects.  This test should
1. download selected projects with expected results from biosimulations.org (e.g. selected projects from biomodels database run by tellerium or copasi).
2. execute corresponding OMEX to generate logs, pdf files and an hdf5 file.
3. compare generated artifacts against expected results from the downloaded projects
4. track compliance by capturing FAULTS in the test (list of known problems).
5. runnable as a unit test with small number of stored projects, or as an integration test with a large number of projects.

PR #904, opened 2023-06-09
Test biosimulations projects

jcschaff:
see #903


jcschaff:
** Summary of work **

- created `BiosimulationsExecTest` which compares vcell executions of biosimulations projects and compares with expected HDF5 using absTol of 1e-9.
  - unit test currently uses BIOMD0000000003, ..06, and ..036 run by tellurium.
  - new python script `downloadProjects.py` to download biosimulations projects with expected results for testing purposes.
- Hack to reduce SEDML supplied ForwardEuler step size by factor of 10 (to pass one of the BSTS tests by falling within stability region).
  - explanation: solver technical parameters depends strongly on scaling ... and vcell computes with concentrations not amounts ... reproducible solver parameters is problematic.
  - we could possibly attempt a complicated analysis to translate the step size - but not worth it.  In practice, no-one would use Forward Euler anyway ... and smaller step size for ForwardEuler is always more accurate unless ridiculously small).
- fixed implementation of Reaction Rates as outputs in OMEX executions
  - `SBMLNonspatialSimResults` now explicitly translates distributed reaction rates to lumped reaction rates for OMEX executions.
  - always coerce lumped kinetics to distributed globally in vcell - but retain the optional flag in some CLI commands.

Issue #918, opened 2023-06-22
replace unsupported tool for notarizing Mac installers


jcschaff:
The tool `xcrun altool` is currently used to notarize Macos installers for the Virtual Cell client. This `altool` was recently unsupported and will stop working in late 2023, and is replaced with the `xcrun notarytool` command.

This issue came to the foreground as we were forced to update the Apple credentials for notarizing the Macos installers using `altool` - but simply replacing the current credentials did not work.  Following the latest instructions from Apple, I was able to authenticate to the notary service using the new tool and available credentials.

`altool` is replaced with `notarytool`.

PR #919, opened 2023-06-22
replace altool with notarytool in Macos installer notarization

jcschaff:
see #918 
use Apple's latest tool `notarytool` to notarize Macos Installers for VCell client, replacing the unsupported and soon to be discontinued `altool`.


jcschaff:
changed a simple constant (time at which simulation warnings will occur), cannot fail tests and within previous range.  Merging prematurely (previous commit passed).

Issue #957, opened 2023-08-28
database backup scripts for new database server


jcschaff:
the vcell-oracle.cam.uchc.edu running Oracle 19c needs to be backed up, not only as a VM, but with archived database dumps using the Data Pipe mechanisms of Oracle.

jcschaff:
Database containers are configured to run with local "New York" timezone - and will follow Daylight savings time as appropriate.  In the future, for more portable services, a more flexible solution cold be installed.

PR #959, opened 2023-08-28
database backup scripts

jcschaff:
see #957

develop, install, and document backup scripts for new database server.


jcschaff:
already merged backup scripts

PR #971, opened 2023-09-13
springsalad xmlproducer

danv61:
Implement the XMLProducer for SpringSaLaD specific attributes / fields
- general attributes
- localized compound spec (species context spec)
- reaction rule specs
Stable, no side effects on other application types


danv61:
committ 5f3a9e1

Issue #973, opened 2023-09-18
Batch export is wrong if "variable reduction" checked


vcellmike:
Batch export does not export eliminated variables if "Variable reduction" button is checked. The example model is Rule-based_Ran_transport_delete25 shared with VCellSupport and danv.

danv61:
Commit 572b339

danv61:
PR #976

PR #976, opened 2023-09-19
Batch simulations only works with 'Variable Reduction' checkbox disabled

danv61:


danv61:
Batch export is wrong if "variable reduction" checked 
Issue #973

PR #979, opened 2023-09-28
Dan ss xml reader

danv61:
The XmlReader and XmlProducer are now complete, export to / import from VCML tested for old application types and for springsalad.


danv61:
Commit 35aa54b

PR #989, opened 2023-10-16
Win11 dll fix

CodeByDrescher:


jcschaff (approved):
works on Windows 11, should be tested on Windows 10.
Approved.

PR #1004, opened 2023-10-27
Misc docker fixes

jcschaff:
supply missing fonts which break some capabilities in VCell CLI container
don't deploy vcell-web service, not used anyway (retain the code until all is transitioned).
remove nuisance JVM property logging in vcell-su admin CLI.
make s3proxy service have same naming convensions as other services (especially in docker-compose).
minor update poetry.lock file


jcschaff:
@CodeByDrescher should fix latest nightly error involving fonts (finger's crossed).

PR #1030, opened 2023-11-11
introduce new vcell-rest module with Quarkus

jcschaff:
vcell-rest will be a REST API with very good OpenAPI support and based on the modern Quarkus framework that supports pluggable Auth and Observability frameworks.

This replaces vcell-api (which has REST endpoints as well as an RPC endpoint, and doesn't support OpenAPI).


jcschaff:
handle SQL injection for publications later.

PR #1041, opened 2023-11-14
Broke down Monolithic BeanUtils into a vareity of more specific utility classes, and added more testing.

CodeByDrescher:
Includes formatting and updating old code to new professional standards.


jcschaff (approved):
Very large number of classes.  I did a reasonably careful review.  It all looks good.

PR #1073, opened 2023-12-07
Recursive dependency linking

CodeByDrescher:
Fixed issues with linux library linking in recent containers


jcschaff (commented):
is this code exercised by a unit test?  The default GitHub runner is a linux box, and the code should be safe for windows and macos (if tests are run there also).

I imagine that a simple unit test which just runs the local native solvers with no inputs and check for bad return code would work.  later we can run that test on multiple runners (windows, mac, linux).

jcschaff (changes requested):
add a test for "running" one (or all) native solvers in a basic way (e.g. no args).  This is a blind spot for us so is needed anyway.  We can chat about the easiest way to test this.

CodeByDrescher:
This is very old, and we currently have no problems described here. Closing

PR #1107, opened 2024-01-18
patch slurm script to use module load singularity/vcell-3.10.0

jcschaff:
patch to production site for slurm submit scripts:

module load singularity/vcell-3.10.0
   instead of
module load singularity


jcschaff:
already deployed

PR #1108, opened 2024-01-22
Added is2D flag, domain name bug fixed in the math

danv61:
Added is2D flag to the UI and internal representation
Note: for now we shoult use it as always false, until we check the solver code and understand its meaning / correctness
domain name bug fixed in the math


danv61:
Issue #1109

PR #1113, opened 2024-01-23
Improve locally stored export metadata

AvocadoMoon:
Human readable export data is written locally within the users home directory where other VCell related data is stored. The human readable data is obtained by getting Simulation data of the current export and using that relevant metadata to produce the human readable export data.


AvocadoMoon:
Double check what happens when handling exports that aren't PDE simulations ensuring that no null error occurs. Also set strings in header to literals.

AvocadoMoon:
Other than those quick reviews, approved by Jim

PR #1117, opened 2024-01-29
Editing or adding a new molecule / species is now properly addressed in existing SS application

danv61:
Changes to molecule / species (done within the biomodel UI) are now properly reflected in the species context spec of an existing application


danv61:
Issue #1118

PR #1129, opened 2024-02-06
Finding spatial data

CodeByDrescher:
Makes clearer error messages when hdf5 results code can not find the needed data


jcschaff (approved):
looks good, approved

PR #1146, opened 2024-02-12
Normalize Time Index

AvocadoMoon:
Instead of having the time name start at the index in the time array, it'll start at 0 and increase from there.


jcschaff (approved):
looks right, approved

Issue #805, opened 2023-02-20
permit VCell client installation without admin privileges


jcschaff:
The agility of the software development process is hampered by the requirement for some end users to have administrative privilege to their machines, thus requiring calls to admin support for some institutions.

VCell client should be made to not require admin privileges if possible.

PR #1166, opened 2024-02-29
Install4J client install as regular user

jcschaff:
see #805


Ezequiel-Valencia:
Already solved by another branch.

PR #1176, opened 2024-03-05
We use the GeometricRegion to recover the correct size of the ROLE_Size parameter

danv61:
We use the GeometricRegion to recover the correct size of the ROLE_Size StructureMappingParameter


danv61:
Issue #1177

PR #1270, opened 2024-06-03
1259 vcell application account gui not responding

AvocadoMoon:


jcschaff (approved):
approved, Zeke agreed to use the exitApplication() function.

Issue #751, opened 2023-01-10
Client Side Annotations Search


abdulmoiz-m:
Create an online search through Bioportal to annotate model elements

danv61:
Dan annotations mix 
PR #1273

danv61:
commit c8437e3

danv61:
PR merged into master

PR #1273, opened 2024-06-06
Dan annotations mix

danv61:
Bringing Search and Add annotations together


danv61:
Issue Client Side Annotations Search #751

PR #1275, opened 2024-06-12
1274 connection creation slash and burn

AvocadoMoon:


jcschaff (approved):
looks great, discussed with Zeke offline.

Issue #1331, opened 2024-07-26
VCell's behavior when closing a window while disconnected is undesirable


CodeByDrescher:
If the client is disconnected (and does not know it's disconnected) when the user tried to close a window with changes, two bad things can occur:
1) the user selects "no" (i don't want to save my changes), and VCell locks up for ~7-30 seconds; no user input registers.
2) the user selects "yes" (I would like to save), and VCell futilely attempts to save the model to a server it hasn't realized it's disconnected from.

This may be related to VCell freezing up overnight, as the logical pathways may be similar.

CodeByDrescher:
Likely fixed, not currently an issue

PR #1332, opened 2024-07-26
Fixed a number of issues, all revolving around VCell locking up under the right (wrong) conditions

CodeByDrescher:
The core issue behind this PR: #1331 

Fixes introduced in addressing this issue:
1) VCell now sports new extensions to the `ClientTaskDispatcher` class:
    a) `ClientTaskDispatcher::dispatchAndWait` will now indefinitely block a dispatched list of tasks, until they are completed. 
    b) `ClientTaskDispatcher::dispatchWithFuture` returns a `Future<Hashtable<String, Object>>` object, which can be used 
          to check if the results are done, or impose a manual timeout on the operation; the underlying mechanism behind 
          `ClientTaskDispatcher::dispatchAndWait`
2) VCell no longer freezes up when disconnected and not trying to save a model in a closing window. 
3) VCell now gives error messages when attempting to save a model while disconnected, informing the user it can't seem to contact the database, and if they proceed a `DataAccessException`. They are no longer stuck behind a progress dialog that goes on for too long. The window will not close until either `No` is re-selected, or the client is reconnected.
4) VCell now realizes much, MUCH sooner it has been disconnected (up to 5 seconds).


CodeByDrescher:
Closes #1331

PR #1374, opened 2024-11-01
Replace unsupported legacy native hdf5 library

jcschaff:
replace use of ncsa.jhdf5 library which is not supported on Mac ARM64 machines with supported HDF5 libraries (io.jhdf and cisd.jhdf5).

- ported spatial simulation HDF5 export **writer** (using cisd.jhdf5)
- ported nonspatial stochastic multi-trial data **reader** (using io.jhdf)
- ported UiTableExporter HDF5 export **writer** (using io.jhdf)


CodeByDrescher (approved):
Looks good to me; nothing too crazy. This seems to just make everything nicer.

PR #1377, opened 2024-11-08
Attempting to enforce needed permissions

CodeByDrescher:


AvocadoMoon (approved):
Seems to ensure file writing is capable, and replace RDF files with empty versions of itself.

PR #1378, opened 2024-11-08
Series of changes to get BSTS to approve of 7.7 Container

CodeByDrescher:
Changes GH actions to use new OIDC method of tokenization


jcschaff (approved):
reviewed along the way and had discussions with Logan, and tested against BSTS.  Approved.

Issue #1336, opened 2024-07-30
Generation of Observables takes a long time 


CodeByDrescher:
When a rule-based model is saved (or flattened to generate a network), the step of generating observables takes disproportionally long time.

danv61:
Duration before:
```
Adding 8 observables to model, 560 ms
removeObsTime: 300,  addParamTime: 229
```
Duration after the fix:
```
Adding 8 observables to model, 20 ms
removeObs Time: 1,  addParam Time: 0
```

danv61:
fixed in 68f793e

PR #1380, opened 2024-11-14
Generation of Observables takes a long time

danv61:
When a rule-based model is saved (or flattened to generate a network), the step of generating observables takes disproportionally long time.
Issue #1336

### DIFF
--- a/vcell-core/src/main/java/cbit/vcell/mapping/NetworkTransformer.java
+++ b/vcell-core/src/main/java/cbit/vcell/mapping/NetworkTransformer.java
@@ -745,9 +745,17 @@ public class NetworkTransformer implements SimContextTransformer {
 		mathMappingCallback.setMessage("generating network: adding observables...");
 		mathMappingCallback.setProgressFraction(progressFractionQuota/8.0f*7.0f);
 		startTime = System.currentTimeMillis();
+
+		long removeObsTime = 0;
+		long addParamTime = 0;
+		long st;		// start
+		long et;		// end
+
+
 		System.out.println("\nObservables :");
 		RbmModelContainer rbmmc = model.getRbmModelContainer();
 		for (int i = 0; i < outputSpec.getObservableGroups().length; i++){
+
 			ObservableGroup o = outputSpec.getObservableGroups()[i];
 //			System.out.println(i+1 + ":\t\t" + o.toString());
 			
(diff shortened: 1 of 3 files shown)


vcell-core/src/main/java/cbit/vcell/mapping/NetworkTransformer.java
```diff
@@ -745,9 +745,19 @@ private void transform(SimulationContext simContext, SimulationContext transform
 		mathMappingCallback.setMessage("generating network: adding observables...");
 		mathMappingCallback.setProgressFraction(progressFractionQuota/8.0f*7.0f);
 		startTime = System.currentTimeMillis();
+
+		long removeObsTime = 0;
+		long addParamTime = 0;
```
jcschaff, line 750: remove this timing code in production code (you can refer to previous commit if you want to find it again).

vcell-core/src/main/java/cbit/vcell/mapping/NetworkTransformer.java
```diff
@@ -781,6 +802,8 @@ private void transform(SimulationContext simContext, SimulationContext transform
 		elapsedTime = endTime - startTime;
 		msg = "Adding " + outputSpec.getObservableGroups().length + " observables to model, " + elapsedTime + " ms";
 		System.out.println(msg);
+		msg = "removeObsTime: " + removeObsTime + ",  addParamTime: " + addParamTime;
+		System.out.println(msg);
```
jcschaff, line 807: remove all System.out or System.err printing ... use Logger instead if you must, or remove if not needed anymore.  Example: lg.debug(msg) when you must have diagnostic messages.

import org.apache.logging.log4j.LogManager; 
import org.apache.logging.log4j.Logger; 
public class NetworkTransformer implements SimContextTransformer { 
      private final static Logger lg = LogManager.getLogger(NetworkTransformer.class);

jcschaff (changes requested):
see comments - generally nice, but lets clean it up.

vcell-core/src/main/java/cbit/vcell/mapping/NetworkTransformer.java
danv61: done

vcell-core/src/main/java/cbit/vcell/mapping/NetworkTransformer.java
danv61: done

vcell-core/src/main/java/cbit/vcell/mapping/NetworkTransformer.java
```diff
@@ -331,24 +335,24 @@ private void transform(SimulationContext simContext, SimulationContext transform
...
+//		msg = "Adding " + outputSpec.getBNGParams().length + " parameters to model, " + elapsedTime + " ms";
+//		lg.info(msg);
 		
 		// ---- Species ------------------------------------------------------------------------------------------------------------
 		mathMappingCallback.setMessage("generating network: adding species...");
 		mathMappingCallback.setProgressFraction(progressFractionQuota/4.0f);
-		startTime = System.currentTimeMillis();
-		System.out.println("\nSpecies :");
+//		startTime = System.currentTimeMillis();
+		lg.info("\nSpecies :");
```
jcschaff, line 355: Minor comment:

1. you may want to remove separate "title" log entries like lg.info("\nSpecies :") and lg.info("\nReactions :") and instead make each subsequent log message more self-contained.  For example, add a "Reaction" prefix to reaction-related log messages if needed.
2. details of processing should use mostly lg.debug() rather than lg.info() to not clog up the logs.  This is especially important during CI/CD testing when such logs make it hard to see the important stuff.

jcschaff (approved):
see comment about logging style - but approved - nice work.

PR #1381, opened 2024-11-14
support Infinity and NaN in SBML values and MathML

jcschaff:
### SBML Import from MathML now supports NaN and Infinity (with MathML unit test)

- `<notanumber/>` MathML imported as new Expression("0/0")
- `<infinity/>` MathML imported as new Expression(Double.MAX_VALUE)

### SBML Import with float values (e.g. value attributes) support NaN and Infinity
- `Double.NaN` => new Expression("0/0")
- `Double.POSITIVE_INFINITY` => new Expression(Double.MAX_VALUE)
- `Double.NEGATIVE_INFINITY` => new Expression(-Double.MAX_VALUE)


CodeByDrescher (approved):
Looks good to me; it's interesting seeing how VCell interprets the needed values!

PR #1382, opened 2024-11-15
Odic deploy fix

CodeByDrescher:
Fixes CD to use ODIC tokens, fixing current and future issues with tokens expiring


jcschaff (approved):
already discussed, good approach, approved.

PR #1383, opened 2024-11-15
Added convenience methods to JhdfUtils

CodeByDrescher:
Adds useful convenience methods for future JHDF use


jcschaff (approved):
looks good

PR #1387, opened 2024-11-19
Redesign the springsalad-specific user interface

danv61:
The large shape depicting the site attributes of the species context spec (springsalad-specific)
Molecular structure properties panel
Tables and table models for site attributes and the links


vcell-client/src/main/java/cbit/vcell/mapping/gui/MolecularTypeSpecsTableModel.java
```diff
@@ -163,45 +190,119 @@ public Object getValueAt(int row, int col) {
...
+			if (aValue instanceof String) {
+				if(sas == null) {
+					sas = new SiteAttributesSpec(scs, mcp, scs.getSpeciesContext().getStructure());
+				}
+				String newExpressionString = (String)aValue;
+				double res = 0.0;
+				try {
+					res = Double.parseDouble(newExpressionString);
+				} catch(NumberFormatException e) {
+					return;
```
jcschaff, line 224: does this behave properly if you type something bad for column_x,y,z?  it seems to ignore things, but it doesn't throw an exception to reject a bad column ... will it just silently not update and not notify the user?

vcell-client/src/main/java/cbit/vcell/mapping/gui/MolecularTypeSpecsTableModel.java
```diff
@@ -163,45 +190,119 @@ public Object getValueAt(int row, int col) {
...
 				if(sas == null) {
-					sas = new SiteAttributesSpec(fieldSpeciesContextSpec, mcp, getSpeciesContextSpec().getSpeciesContext().getStructure());
+					sas = new SiteAttributesSpec(scs, mcp, scs.getSpeciesContext().getStructure());
 				}
 				sas.setDiffusionRate(result);
+				scs.firePropertyChange(SpeciesContextSpec.PROPERTY_NAME_SITE_ATTRIBUTE, null, sas);
+				return;
+			}
+		case COLUMN_COLOR:
+			if (aValue instanceof NamedColor) {
```
jcschaff, line 299: there is now a convenience for downcasting in java, you should probably start using it (more convenient and less verbose).

replace:

if (aValue instanceof NamedColor) {
     NamedColor namedColor = (NamedColor)aValue;
				
with the equivalent statement:

if (aValue instanceof NamedColor namedColor) {

jcschaff (approved):
please look at my comments, overall looks very good.  Approved.

vcell-client/src/main/java/cbit/vcell/mapping/gui/MolecularTypeSpecsTableModel.java
danv61: I figured this will fail as a result of a typo, most likely the user will instantly figure out what went wrong. I didn't want to bother him with an exception dialog he'll have to close

vcell-client/src/main/java/cbit/vcell/mapping/gui/MolecularTypeSpecsTableModel.java
danv61: I know about it, you showed me, but old habits die hard... I will try to remember to use it

PR #1390, opened 2024-11-21
use Tracer for additional observability into OMEX execution failures

jcschaff:
- integrated the Tracer mechanism into CLIRecordable to give better observability into OMEX processing errors.
- verified that many BIOMD000XXX.omex files passed OMEX execution testing now which had failed due to NaN, Infinity, and Python processing (see fault list in BSTSBasedOmexExecTest.java)
- verified that many more BIOMD000XXX.xml SBML files passed SBML import testing (see fault list in SBMLTestSuiteTest.java)
- created BiosimulationsLog java classes for JSON I/O for biosimulations structured logging (not yet integrated).  The idea is that we can create this log file within java rather than sending a large number of calls to ask Python to write the logs for us.


CodeByDrescher (approved):
Went through with Jim, all looks as expected, and some neat features added I can't wait to see developed further!

PR #1391, opened 2024-11-21
Links are now editable

danv61:
Links are now editable, coordinates will be recalculated
Cosmetic fixes, rendering precision for coordinates and links rendering truncated to angstrom (one decimal unit)
In the table cell renderer the ColumType enum is recovered by column label instead of by column index (because some columns might have been moved around or removed at runtime)

dan-ss-links branch


jcschaff (approved):
looks great Dan, only comment is to remove commented out code unless you REALLY think that you'll use it soon.  Approved.

PR #1392, opened 2024-11-21
OMEX Execution fails with repeated task of size 1

jcschaff:
OMEX Execution fails with repeated task of size 1

This effected BMDB models 561,679,680,681,684,724,915

All but 561 now pass.  561 fails because there is nothing to solve - no ODEs - and VCell solvers get confused.


CodeByDrescher (approved):
It's basically 1 simple addition and a reorganization for clarity; nothing crazy; Approved

PR #1393, opened 2024-11-22
Updated KiSAO OBO and added parsing for KiSAO_694

CodeByDrescher:


vcell-core/src/main/java/cbit/util/kisao/KisaoTermParser.java
```diff
@@ -22,16 +24,15 @@ public class KisaoTermParser {
...
-		InputStream is = KisaoTermParser.class.getClassLoader().getResourceAsStream(Kisao_OBO);
-		BufferedReader isr = new BufferedReader(new InputStreamReader(is));
-		String line = null;
-		boolean inState = false;
 		KisaoOntology ontology = new KisaoOntology();
 		KisaoTerm curr = null;
+		boolean inState = false;
+
+		InputStream is = KisaoTermParser.class.getClassLoader().getResourceAsStream(Kisao_OBO);
+        assert is != null;
```
jcschaff, line 32: assert's are not great in production code - just throw some sort of IOException or Runtime (assert's are not designed to be handled in production).

jcschaff (approved):
style wise - the assert dealing with a misconfiguration should be replace with some other exception.  asserts are best used in test code.

Approved anyway.

PR #1395, opened 2024-12-06
request focus for VCell client after OIDC login/logout

jcschaff:
request focus for VCell client after OIDC login/logout.  This may not be supported or may be blocked by security on some platforms - but it is harmless.


AvocadoMoon (commented):
Seems straight forward, ask the computer if it allows application foreground requests. If it does, then request focus and if not continue the flow as usual. My only feedback would be to change the error logging to a warning if the client does not support application focus changing.

AvocadoMoon (approved):
Everything seems perfect now, allowing supported devices to use the foreground redirection.

PR #1396, opened 2024-12-08
write Biosimulation log file directly from Java

jcschaff:
### Background
Biosimulators are required to write structured status/log YAML files when executing OMEX files according to a particular format.  biosimulator-vcell currently generates and updates this file by frequent Python calls to wrapper.py and status.py in vcell-cli-util in order to reuse some existing code.  However this approach of using Python to write and update these files is inefficient and obscures our understanding of status throughout an OMEX execution.  We want to track status and write the Biosimulation log file directly from Java.

### Previous work
As part of PR #1390, the BiosimulationLog classes were created and tested to provide a Java binding to this log file structure which can be serialized and deserialized in JSON or YAML.

### This PR
In this PR, the calls to PythonCalls in vcell-cli which invoked status.py were reimplemented in Java using BiosimulationLog.java, and status.py was removed.  Now, only OMEX file validation and plot generation require Python processing.


CodeByDrescher (approved):
Passes log tests in BSTS; approved!

PR #1397, opened 2024-12-10
Filter springsalad simulation results

danv61:
Filter the results of running a springsalad simulation.
The results are solver generated and are presented as linear combinations of each molecule's site bond / state pair
Also we highlight trivial result sets (where the data doesn't change)
branch: dan-ss-filter-results


jcschaff (approved):
I spoke with Dan about the introduction of .isSpringSaLaD() in the data layer, and he promised to remove these later in the process (future PR).

Approved.

PR #1398, opened 2024-12-11
redefine simulation JobIndex in terms of ScanIndex and TrialIndex

jcschaff:
Introduced `ScanIndex` class to define parameter overrides indexing in a type safe manner.
Introduced `TrialIndex` class to define <no-yet-implemented> simulation trials indexing in a type safe manner.
`jobIndex` is still just an integer and is mapped to ScanIndex and TrialIndex.

`MathOverrides` owns **Scan Count** and maps between `ScanIndex` and parameter override values in a scan.
`SolverTaskDescription` owns **Number of Trials** and seeds to stochastic trials (Trialindex is always 0 for this PR).
`Simulation` maps between jobIndex, ScanIndex and TrialIndex.

As of this PR, if num trials is > 1, then each trial is identical.

The current mapping is jobIndex = trial_index * scanCount + scanIndex, but please use utility functions in Simulation to map between these indices.  This rasterization will perform complete scans for each trial, so that during execution, each trial is executed in order.


danv61 (approved):
Looks very good.

PR #1400, opened 2024-12-13
Fast in-memory biosim logging

jcschaff:
- `BiosimulationLog` is now a singleton stored with an init(), close() lifecycle and stores the log output directory and the `ArchiveLog` object.
- The `ArchiveLog` object encodes the hierarchical structure of the logs and is one-to-one with the JSON/YAML.  All logging merely updates the `ArchiveLog` data structure in memory until it is written to the `log.yml` file during close().
- Online validation is disabled by default but could be manually enabled via `BiosimulationLog.VALIDATE_LOG` boolean where it would be invoked upon close().
- For debugging purposes one may view the in-memory log data structure `BiosimulationLog.instance().archiveLog`, or if needed, one may temporarily add a call to `BiosimulationLog.write()` inside of `BiosimulationLog.setDirty()`.


CodeByDrescher (approved):
There will be further small type changes, but since java is strongly typed, I don't think further review will be necessary (unless other semantic changes are needed).

Looks like a well made, great speed-up, and cleanup to boot!

PR #1401, opened 2024-12-13
Improvements to springsalad specific properties panel and shape representation

danv61:
selected shape highlight - site oval shape and link line shape
interactive selection updates between the sites / links table and the corresponding shapes - works both ways

### DIFF
--- a/vcell-client/src/main/java/cbit/vcell/mapping/gui/LinkSpecsTableModel.java
+++ b/vcell-client/src/main/java/cbit/vcell/mapping/gui/LinkSpecsTableModel.java
@@ -149,12 +149,6 @@ public class LinkSpecsTableModel extends VCellSortTableModel<MolecularInternalLi
              * to, or greater than the second.<p>
              */
             public int compare(MolecularInternalLinkSpec mils1, MolecularInternalLinkSpec mils2) {
-
-
-
-
-
-
                 ColumnType columnType = columns.get(col);
                 switch (columnType) {
                     case COLUMN_LINK:
--- a/vcell-client/src/main/java/cbit/vcell/mapping/gui/LinkSpecsTableModel.java
+++ b/vcell-client/src/main/java/cbit/vcell/mapping/gui/LinkSpecsTableModel.java
@@ -150,6 +150,11 @@ public class LinkSpecsTableModel extends VCellSortTableModel<MolecularInternalLi
              */
             public int compare(MolecularInternalLinkSpec mils1, MolecularInternalLinkSpec mils2) {
 
+
+
+
+
+
                 ColumnType columnType = columns.get(col);
                 switch (columnType) {
                     case COLUMN_LINK:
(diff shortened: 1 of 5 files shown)


vcell-client/src/main/java/cbit/vcell/mapping/gui/LinkSpecsTableModel.java
```diff
@@ -260,6 +259,24 @@ protected List<MolecularInternalLinkSpec> computeData() {
...
             return null;
         }
+
+        Collections.sort(allParameterList, new Comparator<MolecularInternalLinkSpec>() {
+            @Override
+            public int compare(MolecularInternalLinkSpec mils1, MolecularInternalLinkSpec mils2) {
+                SpeciesContextSpec scs = mils1.getSpeciesContextSpec();
+                if(fieldSpeciesContextSpec != scs) {
+                    throw new RuntimeException("SpeciesContextSpec inconsistent.");
+                }
```
jcschaff, line 269: do you need to verify that mils2 also points to the correct SpeciesContextSpec?  it is okay if you think that is overkill.

vcell-client/src/main/java/cbit/vcell/mapping/gui/MolecularStructuresPropertiesPanel.java
```diff
@@ -140,11 +166,39 @@ public boolean isViewSingleRow() {
                 }
             };
             shapePanel.addMouseMotionListener(new MouseMotionAdapter() {
+                @Override
                 public void mouseMoved(MouseEvent e) {
+                    // this will fire if the mouse moves over the panel, we may need it for drag and drop
+                    super.mouseMoved(e);
+//                    Point overWhat = e.getPoint();
```
jcschaff, line 173: remove when you are done with this code.

vcell-core/src/main/java/cbit/vcell/mapping/SpeciesContextSpec.java
```diff
@@ -70,8 +70,14 @@ public class SpeciesContextSpec implements Matchable, ScopedSymbolTable, Seriali
...
     private static final String PROPERTY_NAME_WELL_MIXED = "wellMixed";
     private static final String PROPERTY_NAME_FORCECONTINUOUS = "forceContinuous";
-    public static final String PROPERTY_NAME_SITE_ATTRIBUTE = "SiteAttributes";     // springsalad
-    public static final String PROPERTY_NAME_LINK_LENGTH = "LinkLength";            // springsalad
+    // springsalad group
+    public static final String PROPERTY_NAME_SITE_ATTRIBUTE = "SiteAttributes";
+    public static final String PROPERTY_NAME_LINK_LENGTH = "LinkLength";
+    public static final String PROPERTY_NAME_SITE_SELECTED_IN_TABLE = "SiteSelectedInTable";
+    public static final String PROPERTY_NAME_LINK_SELECTED_IN_TABLE = "LinkSelectedInTable";
+    public static final String PROPERTY_NAME_SITE_SELECTED_IN_SHAPE = "SiteSelectedInShape";
```
jcschaff, line 78: for these 4 new bound properties, make sure you un-listen and re-listen where appropriate to (1) keep the UI responsive even after add/remove of objects and (2) prevent memory leaks and invisible components trying to update themselves.

jcschaff (approved):
Please look at comments and address them as appropriate (if at all).  Looks very nice and fun Graphics code.

vcell-client/src/main/java/cbit/vcell/mapping/gui/LinkSpecsTableModel.java
danv61: Actually even the test for mils1 is overkill, I'm only trying to catch bugs early during development. Condition should bever happen in a bug-free code.

vcell-client/src/main/java/cbit/vcell/mapping/gui/MolecularStructuresPropertiesPanel.java
danv61: I'll probably remove the whole mouse motion listener eventually, unless it's needed for the drag and drop implementation

PR #1402, opened 2024-12-16
unify fault reporting across unit and nightly tests for Omex execution

jcschaff:
- OmexTestCase objects hold test cases with expected results - stored in a newline-delimited JSON file (`test_cases.ndjson`)
- The BSTSBasedOmexExecTest use the text cases defined in `test_cases.ndjson` as source of truth for regression testing of OMEX executions.
- A new CLI command (execute-omex (copied/simplified Nightly tests to use test_cases.ndjson as source of truth for regression testing against all BIOMD OMEX files.
- OmexTestingDatabase unifies runtime error identification between both execute-omex CLI command and the BSTS based OMEX unit tests. 
  - has 'determineFault()' and 'summarize()' methods and a placeholder 'generateReport()' method which will generate a nice error report in the future - **via a new `omex-report `cli command.**
- introduces new 'execute-omex' cli command (derived from 'execute' command, for single omex files with new reporting)
  - each run generates an 'exec_summary.json' file with a single OmexExecSummary object - to be appended to a master 'exec_summary.ndjson' file (one report per line).  These `OmexExecSummary` objects are to be compared with the list of OmexTestCase objects from `test_cases.ndjson` for final reporting.
  - it also generates another json file per run, a 'tracer.json' file which holds all tracer events ... we can output this in appropriate formats, or pretty print it for human use.  the `tracer.json` file has all of the information from the logs generated by `CLIRecordable`... but in a structured format (`EventLog` objects with context) which is machine understandable.


CodeByDrescher (approved):
Really like this! SUPER intuitive, and a very useful feature-set!

jcschaff:
@CodeByDrescher thanks for the review.

PR #1404, opened 2024-12-23
Fixed BMDB Nightly testing, and upgrade with better logging

CodeByDrescher:


vcell-cli/src/test/java/org/vcell/cli/testsupport/TestOmexTestReport.java
```diff
@@ -0,0 +1,33 @@
+package org.vcell.cli.testsupport;
+
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+import org.vcell.sedml.testsupport.*;
+
+public class TestOmexTestReport {
```
jcschaff, line 7: might as well stick in expected results of some kind - or expectThrow if row,column out of range ... or something.

Also, this needs a test category, like:
@Tag("Fast")

jcschaff (approved):
looks good, left comment about unit test, won't run in CI because no @tag() annotation, also, doesn't test for expected results or expected failures.

add tag to unit test or it won't run.

Approved - nice!

vcell-cli/src/test/java/org/vcell/cli/testsupport/TestOmexTestReport.java
CodeByDrescher: Had the result checked to see if it matched the expected.

PR #1409, opened 2025-01-08
Help generate docs

AvocadoMoon:


jcschaff (approved):
Zeke and I paired on this one during review, based on Zeke's earlier PR.  Approved.

PR #1412, opened 2025-01-09
Sync Without Attempting To Keep Origins Permissions

AvocadoMoon:
Instead of using the flag which attempts to keep source file permissions, let them be whatever the target file permissions are, updating the target files time stamp in the process.


jcschaff (approved):
I see the benefit of this change, and seems low risk - approved.

PR #1413, opened 2025-01-09
Dan ss random seed2

danv61:
Add SIMULATION OPTIONS block to the solver input file
Add optional RandomSeed field to the block


danv61:
commit 75130bd

CodeByDrescher (approved):
The CI failure is due to old master fork, where master was failing; shouldn't be a problem.

PR #1414, opened 2025-01-09
Synchronize old "Annotations" with Formal annotations in Sims

danv61:
Synchronize old "Annotations" with Formal annotations in Sims 
Issue #1408
Replaced Annotation with Description in SimulationSummaryPanel, since Annotations have their own panel
Part of hackaton2, Jan 2025


AvocadoMoon (approved):
Simple change of label to be more accurate to it's representation. Looks good.

PR #1415, opened 2025-01-10
new users-query CLI command lists users who ran sim since a start date

jcschaff:
administrative tool to query the list of users who ran simulations since a given date.  user list includes email address, name, userid and whether they agreed to be notified.


AvocadoMoon (approved):
Everything seems good. The command line execution, and writing of results seems reasonable. Only portion that may be of concern is the SQL because it's a lengthy query command leaving room for error, but reading it everything seems okay.

PR #1419, opened 2025-01-14
User authentication middleware

AvocadoMoon:
Make it such that there is no VCell guest used in the new API servers, and individual endpoints don't need to verify if an anonymous user is attempting access.


vcell-rest/src/main/java/org/vcell/restq/db/UserRestDB.java
```diff
@@ -42,22 +43,34 @@ public UserRestDB(AgroalConnectionFactory agroalConnectionFactory) throws DataAc
 
 
     // TODO: add some short-lived caching here to avoid repeated database calls
-    public User getUserFromIdentity(SecurityIdentity securityIdentity, boolean defaultGuest) throws DataAccessException {
+
+    /**
+     * Get the users identity from the requests that they make, usually this is a JWT token, placed within the
+     * HTTP Authorization header. If defaultGuest is true, then null will be returned for anonymous users and guests, otherwise
+     * and error will be thrown about the user being unauthenticated.
+     * Tokens generated by the old API for guest users will have no effect on this function because the securityIdentity
```
jcschaff, line 51: remove any notion of a guest user please ... on an endpoint by endpoint basis, one may pass down a DUMMY_USER, but this should not be a global strategy.  Really this is because we don't have roles and authorization for writing - other

vcell-rest/src/test/java/org/vcell/restq/apiclient/UsersApiTest.java
```diff
@@ -164,6 +170,11 @@ public void testOldAPITokenGeneration() throws ApiException {
         Assertions.assertEquals("Administrator", token.getUserId());
     }
 
+    /**
+     * If there is no user mapping for the client or the user does not have an JWT token in the Authorization header for HTTP,
+     * throw 401. If the user is a guest, return a token with the user id "vcellguest" and the user key "140220477".
```
jcschaff, line 175: "if the user is a guest" is misleading to me because it implies that with respect to the UsersApi, the user identified itself as a "guest", but rather the usersResourceApi.getGuestLegacyToken() should not require authentication at all.

vcell-rest/src/test/java/org/vcell/restq/apiclient/UsersApiTest.java
```diff
@@ -178,6 +189,27 @@ public void testOldAPITokenGenerationForGuest() throws ApiException {
         Assertions.assertEquals("140220477", guestApiToken.getUserKey());
     }
 
+    @Test
+    public void testUserMiddleWare() throws ApiException{
+        ApiClient defaultUser = TestEndpointUtils.createUnAuthenticatedAPIClient(testPort);
```
jcschaff, line 194: not "defaultUser" but anonymous user.

vcell-rest/src/test/java/org/vcell/restq/apiclient/UsersApiTest.java
```diff
@@ -178,6 +189,27 @@ public void testOldAPITokenGenerationForGuest() throws ApiException {
...
+        PublicationResourceApi publicationResourceApi = new PublicationResourceApi(defaultUser);
+
+        // public available to everyone
+        Assertions.assertDoesNotThrow(publicationResourceApi::getPublications);
+
+        // guests can not delete a publication
```
jcschaff, line 200: anonymous users (or unauthenticated calls) can not delete a publication.  Note that there are additional roles necessary for deleting publications.

jcschaff (changes requested):
please address comments.

vcell-rest/src/test/java/org/vcell/restq/apiclient/UsersApiTest.java
```diff
@@ -153,8 +155,12 @@ public void testOldAPITokenGeneration() throws ApiException {
         Assertions.assertEquals ("vcellNagios", token.getUserId());
 
         // Bob requests
-        token = bobUserApi.getLegacyApiToken();
-        assert (token.getToken() == null);
+        try{
+            bobUserApi.getLegacyApiToken();
+            Assertions.fail("Should throw 401 since only clients with role user can call it.");
```
jcschaff, line 160: maybe change message that "should throw 401 since only users mapped to VCell Users can authenticate to this endpoint"

User bob does have role "user" it is just not mapped to a vcell user.
%test.quarkus.keycloak.devservices.users.bob=bob
%test.quarkus.keycloak.devservices.roles.bob=user

jcschaff (approved):
looks good, please address my comments before merging.

vcell-rest/src/main/java/org/vcell/restq/handlers/FieldData/FieldDataResource.java
```diff
@@ -99,6 +99,7 @@ public FieldDataShape getFieldDataShapeFromID(@PathParam("fieldDataID") String f
     @Operation(operationId = "analyzeFieldDataFile", summary = "Analyze the field data from the uploaded file. Filenames must be lowercase alphanumeric, and can contain underscores.")
     public AnalyzedResultsFromFieldData analyzeFieldData(@RestForm File file, @RestForm String fileName){
         try{
+            userRestDB.getUserFromIdentity(securityIdentity, UserRestDB.UserRequirement.REQUIRE_USER);
```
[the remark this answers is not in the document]
AvocadoMoon: Talked it out with Jim. File handling can be a security nightmare, so forcing a user to be tied to the file being uploaded can be used to deter bad actors.

PR #1420, opened 2025-01-14
 Stricter and better integrator filtering for some application types.

danv61:
Refactoring. Moved the SolverFeatureSet name inside the class.
Dismantling the Map since it was duplicating existing ArrayList


CodeByDrescher (approved):
Looks good, and makes sense

PR #1421, opened 2025-01-16
don't save trivial numTrials value to XML and VCML

danv61:
don't save trivial numTrials value to XML and VCML (default value is 1)


AvocadoMoon (approved):
Seems to only allow changes when the number of trials is greater than 1. Everything looks good.

PR #1422, opened 2025-01-16
updating tests

CodeByDrescher:
Updating two tests to reflect actual results


jcschaff (approved):
BIOMD 775 now fails when processing 
`<apply>
     <csymbol encoding=\"text\" definitionURL=\"http://www.sbml.org/sbml/symbols/rateOf\"> rateOf </csymbol> 
     <ci> G </ci>
</apply>`.  
This is clearly not supported/implemented in VCell - the model is asking for the time derivative of G within MathML.  I'm not sure why this ever passed - maybe it was previously skipped?


BIOMD 956 fails in Python while parsing a SEDML document during plotting - this will no longer be needed when the new Java-based plotting is introduced.  The original SEDML is parsed without error earlier in the run.

PR #1424, opened 2025-01-21
Dan num trials core

danv61:
Adding numTrials to solver invocation and result display logic
Insuring backward compatibility with older applications spatial / non-spatial stochastic / deterministic
Testing parameter scan for multiple solvers - Smoldyn, NFSim, Gibson, aso


jcschaff (approved):
reviewed with Dan and suggested a few small fixes, looks good, Approved.

PR #1425, opened 2025-01-22
Client Server Refactor

AvocadoMoon:
Refactor the client, server architecture such that each is independent and making changes on the client can't influence changes on the server.


vcell-client/src/main/java/org/vcell/util/network/RemoteRegistrationService.java
```diff
@@ -1,13 +1,14 @@
-package org.vcell.service.registration.remote;
+package org.vcell.util.network;
```
jcschaff, line 2: can we just remove this entire class? .. we do registration in Auth0 and in vcell-rest

jcschaff (changes requested):
very nice job, please take a look at my comments, especially test VCellApiDevClient and its ability to log in.

vcell-client/src/main/java/org/vcell/util/network/RemoteRegistrationService.java
AvocadoMoon: In line 52 of the VCell Client Module it states how the registration service interface is not clean, and that it mixes new registration with updates. Idealistically this class should be capable of removal, however, I'm not sure if it should be and the time to review it may be substantial.

jcschaff: okay, don't remove it in this PR

jcschaff (approved):
changes to accommodate VCellClientDevMain look good - very nice job on this PR, Approved!

PR #1426, opened 2025-01-22
add quarkus filter to sort HTTP response codes in Openapi spec

jcschaff:
The OpenAPI spec generated from Quarkus had often randomly permuted the HTTP response codes, creating non-deterministic noise in the OpenAPI spec and the generated API client code.

A filter was added to the openapi generation step in Quarkus to force the order of the APIResponses to be alphabetical (e.g. "200", "401", "403")


AvocadoMoon (approved):
Looks good. It takes the API spec during build process and sorts it appropriately.

PR #1427, opened 2025-01-23
Plotting in Java for VCell CLI (+ Python live shell "removal" & deprecation)

CodeByDrescher:
Plotting the data from in CLI had been handled by calling python functions through a python shell generated in java. 

This approach had caused a number of problems, from speed to debugging difficulties, etc.

This PR deprecates the live-python client, and removes all calls to it from CLI. Instead, a java-based plotting library has been added, and tests included to ensure no functionality has been lost or damaged in the conversion process.


jcschaff (approved):
nice job, reviewed this interactively with Logan, suggested comments for logging, which was already remedied.  Approved!

PR #1431, opened 2025-01-29
Sbml to finite volume input

AvocadoMoon:
Add an endpoint which allows for SBML files users have to be transformed into finite volume solver files.


jcschaff (approved):
spoke with @AvocadoMoon about endpoint names and removing the zip file and content files before endpoint returns.

PR #1432, opened 2025-01-31
Fix Outdated Artifact Function Being Used

AvocadoMoon:
Change the artifact version from v3 to v4.


CodeByDrescher (approved):
Name changes required with v4. Looks simple and solid

Issue #1434, opened 2025-02-04
Trying to open one's own uncurated public model fails, and thereafter opening any model fails


ACowan0105:
To duplicate:
Change permissions on one of your own models to "Public" .  The model will now appear in the Public>Uncurated folder.  Try to open the model from there.  It will fail to open and in the dialog box "open" will be unavailable.  Go back to any other model (public or private).  Now, you might not be able to open any other model.

danv61:
PR #1435 Enable opening of any biomodel unless it's a duplicated leaf 
Fixed in 5ef5de7

danv61:
Fixed for both BioModels and MathModels.
Merged into master.

PR #1435, opened 2025-02-04
Enable opening of any biomodel unless it's a duplicated leaf

danv61:
Issue #1434
Trying to open one's own uncurated public model fails, and thereafter opening any model fails 
Part of hackathon Feb 04 2025


CodeByDrescher (approved):
Seems solid to me

PR #1437, opened 2025-02-04
Fixed bug with cropping in geometry viewer

### LINKED ISSUE #1436: Cropping a geometry results in an improper extent
Looks like the whole image is scaled by area rather than being properly scaled by each new extent in x and y

CodeByDrescher:
closes #1436


danv61 (approved):
Looks good. Approved.

PR #1439, opened 2025-02-10
Langevin solver multiple runs post processing

danv61:
Branch: dan-ss-post-processing
Calculate trivial statistics for multiple runs: averages, standard deviation, min, max
Proof of concept / demo for displaying these statistics graphically
for solvers-langevin-mac.version we use the amd64 (intel)  (13) - see the pom.xml changes (3 files)


jcschaff (approved):
reviewed the code with @danv61 and suggested changes which were incorporated.  Also, we just discussed the macOS amd64 vs arm64 issue for the langevin solver - the decision is to use the amd64 solver on all macOS platforms.

Approved.

PR #1442, opened 2025-02-17
Change Endpoint From 'Get' to 'Post'

AvocadoMoon:
The generated python client was not submitting files to the finite volume solver endpoint because the endpoint required the header 'GET'. Changed the header to 'POST' and now files are being submitted.


jcschaff (approved):
good catch, approved.

PR #1443, opened 2025-02-17
Fix sbml debug log error

AvocadoMoon:
JSBML in debug mode throws errors so return it to INFO.


jcschaff (approved):
Good.  avoid and document JSBML debug logging runtime errors - Approved!

PR #1444, opened 2025-02-24
As required by BioSimulations, plots reported as Reports

CodeByDrescher:
BioSimulations requires simulators to output plot data as though it was a report, so that it can easily plot data on it's website.

In the process of making this happen, we've overhauled how CLI passes data to the BioSim HDF5 writer class, streamlining the whole process, and removing the use of intermediary HDF5 in spatial simulations. Simulations should now run much faster for spatial sims.


danv61 (approved):
Looks well organized. 
Approved.

Please add three to Triplet.toString() and hashCode() before merge

PR #1445, opened 2025-02-24
add API endpoint to generate fv solver input from vcml

jcschaff:
There is now 2 endpoints to generate FiniteVolume input files:
1. added `/api/v1/solver/getFVSolverInputFromVCML` endpoint (which accepts a VCML file and a spatial simulation)

2. extended the existing endpoint `/api/v1/solver/getFVSolverInput` (which accepts SBML Spatial) to add parameters for duration and output time interval).


AvocadoMoon (approved):
Everything looks good. Adding an endpoint which converts VCML to FiniteVolume input.

Issue #1453, opened 2025-03-05
Langevin input file gets wrong particle location name


danv61:
we can't use the "Extracellular_Intracellular_membrane" math name of the subdomain,
the Langevin solver has no clue about what that might be
luckily we enforce that a membrane molecule must be anchored to exactly one membrane (named "Membrane"), we use that

Replace
            `MOLECULE: "Anchored" Extracellular_Intracellular_membrane Number 21 Site_Types 5 Total_Sites 5 Total_Links 4 is2D false`
with
            `MOLECULE: "Anchored" Membrane Number 21 Site_Types 5 Total_Sites 5 Total_Links 4 is2D false`

danv61:
Fixed in commit 0cefb78ef21201be681353dcbd029203c41acaa7

danv61:
Merged into master

PR #1454, opened 2025-03-05
Langevin input file gets wrong particle location name

danv61:
Issue #1453, Langevin input file gets wrong particle location name
better issue message for case when Kf exceeds the diffusion limited rate.
calculate binding reaction radius for all sites


CodeByDrescher (approved):
Looks good to me. Approved

PR #1455, opened 2025-03-11
prepare for nativelib deployment through libvcell repo

jcschaff:
remove vcell-nativelib (moved to libvcell)
update any dependencies to make native code work well on multiple platforms.


AvocadoMoon (approved):
Everything seems good. Replaced problematic library JDOM to a newer version capable of native compilation, and made copying geometry easier.

PR #1457, opened 2025-03-14
Field Data From Simulation

AvocadoMoon:
Fix the bug which causes field data creation from simulations to fail. Also port over said functionality to the new REST API.


jcschaff (approved):
looks good - please merge and deploy to Alpha when the maven repo problem resolves itself - make sure it passes all tests.

PR #1458, opened 2025-03-14
Admin docs

jcschaff:
Documented the new VCell Oracle Database users 'vcell_service' and 'vcell_dev' which have role 'vcell_role' (and a trigger) to give access to VCELL schema.

Also, updated the manual instructions for website TLS cert (not Let's Encrypt) - thanks to Mike Wilson for doing the work.


AvocadoMoon (approved):
Sensible documentation for adding more users to the DB and updating certificates.

PR #1463, opened 2025-03-24
Copy Field Data From BioModel

AvocadoMoon:
Add an endpoint that allows for copying all field data contained within a BioModel, and make it be only on the server side instead of client control.


jcschaff (approved):
approved - reviewed with Zeke and suggested changes which were incorporated.

Issue #1433, opened 2025-02-04
Race condition when first displaying results after running simulations or new application - Windows only


vcellmike:
Two ways to get an error:

- Create a reaction, set non-zero parameters, run before being saved.
- Copy application as a new, run before saving.

Error:

RuntimeException-Failed to determine metadata for data symbols: Cannot assign field "color" because "this.root" is null
NullPointerException-Cannot assign field "color" because "this.root" is null

danv61:
Fixed (hopefully) in 65d57eb

danv61:
PR #1465 NullPointerException in TreeMap because of concurrent modification

danv61:
Merged into master

PR #1465, opened 2025-03-26
NullPointerException in TreeMap because of concurrent modification

danv61:
NullPointerException in TreeMap because of concurrent modification
Replace use of TreeMap with ConcurrentSkipListMap which is thread safe


danv61:
Fixes Issue #1433 Race condition when first displaying results after running simulations or new application - Windows only

CodeByDrescher (approved):
Solid replacement choice, approved

PR #1466, opened 2025-03-26
allow overriding field data writing in FiniteVolumeFileWriter

jcschaff:
needed by libvcell to create a custom LocalFiniteVolumeFileWriter which can customize local field data usage


CodeByDrescher (approved):
Understandable reason and modification; approved.

PR #1467, opened 2025-03-26
Changes to VCell to support migration to new Slurm cluster

jcschaff:
- make ssh cmd options configurable via properties from K8s config (ControlPath option needed tweaking)
  -  K8s ConfigMaps and Dockerfiles use env var `vcell_ssh_cmd_options_csv`
  - Java code uses property `vcell.ssh.cmd.options.csv` 
  - default value: `StrictHostKeyChecking=No,ControlMaster=auto,ControlPath=~/.ssh/%r@%h:%p,ControlPersist=1m`
  - corresponding ssh option syntax: `-o StrictHostKeyChecking=No -o ControlMaster=auto -o ControlPath=~/.ssh/%r@%h:%p -o ControlPersist=1m`

### DIFF
--- a/vcell-core/src/main/java/cbit/vcell/message/server/cmd/CommandServiceSshNative.java
+++ b/vcell-core/src/main/java/cbit/vcell/message/server/cmd/CommandServiceSshNative.java
@@ -1,21 +1,13 @@
 package cbit.vcell.message.server.cmd;
 
+import cbit.vcell.resource.PropertyLoader;
+import org.vcell.util.exe.ExecutableException;
+
 import java.io.File;
 import java.io.IOException;
 import java.io.PrintWriter;
 import java.io.StringWriter;
-import java.util.ArrayList;
-import java.util.Arrays;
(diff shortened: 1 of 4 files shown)


vcell-core/src/main/java/cbit/vcell/message/server/cmd/CommandServiceSshNative.java
CodeByDrescher, line 1: Note the auto-condensation of the individual util import into a glob import on line 10; I think I remember you saying you preferred to undo that.

CodeByDrescher (approved):
Looks solid, approved

vcell-core/src/main/java/cbit/vcell/message/server/cmd/CommandServiceSshNative.java
jcschaff: good catch - I had struggled with that one a bit, but decided that it was okay for java.util classes in this case - also the automated "organize imports" kept applying it.

PR #1471, opened 2025-04-07
VCell Connection Failure Fix

AvocadoMoon:
### Problem
After extended period of time VCell would be unable to create a new connection to the servers, and perform functions.

### Solution
Fix the refresh system within the Java client and Auth0, such that now refresh tokens function appropriately.


AvocadoMoon:
Modified the Java client, Angular app, and Auth0 config.

jcschaff:
@AvocadoMoon stepped me though the recent Angular security changes - looks good.

PR #1478, opened 2025-04-11
Site physical limits

danv61:
Checking for physical limitations - site size, diffusion rate, partition size


danv61:
Issue #1479
Site physical limits - site radius, diffusion rate vs partition size

CodeByDrescher (approved):
Looks good to me, approved

PR #1484, opened 2025-04-22
Remove Field Data Specs

Ezequiel-Valencia:
Field data was accessed using these Java objects that could be multiplexed into different states of execution and results. Now each execution and result are their own object simplifying the flow for each functionality of field data.


Ezequiel-Valencia:
In addition to removing field data spec from the VCell repository I made the error handling improve for the communication layer in both Quarkus and Java client.

CodeByDrescher (approved):
Took a general look for any signs of broken pipes, or other data loss. Looks solid; further PR should be done for algorithmic analysis, in parts where non-op refactor was not the primary goal.

PR #1490, opened 2025-04-28
Simple field data endpoint

Ezequiel-Valencia:
An endpoint where you upload a file, and field data is created.


CodeByDrescher (approved):
Semantically looks good, however I can't entirely vouch for openapi files, as I have very limited experience in that area.

PR #1496, opened 2025-05-06
Update BMDB model list - supported / failed

danv61:
BMDB models BIOMD0000000355 and BIOMD0000000355 fail to import in vcell because some compartments size are marked as variable, which is unsupported in vcell


danv61:
Code to go through all the BMDB models and check if SBML import succeeds
Make lists with new failures, make list with models marked as failures but now csn be imported.

jcschaff (approved):
Dan, looks good.   Approved!

If you haven't already, please make a follow-up issue to run the SBML import against all BMDB models periodically (e.g. every week) and make a discrepancy report to send to slack.

PR #1498, opened 2025-05-07
Use Exsiting VCell Connection for Parmater Estimation Request

Ezequiel-Valencia:
Instead of creating a request for every estimation the user does, utilize the existing connection to skip re-authentication.


jcschaff (approved):
looks good, changes made via discussion.

PR #1499, opened 2025-05-08
Field data hot fix

Ezequiel-Valencia:
Fixed the following bugs remaining with field data:

- Creation using a file didn't work on the Java client because the file was not being uploaded
- Error handling on the server did not delete database entries when creation failed
- When the user object was returned with an external data identifier, it did not have enough information to act as a User object on the client


Ezequiel-Valencia:
Suggested Changes:

- [ ] Break up the error handling for saving field data

jcschaff (approved):
approved if comment for breaking up error handling is incorporated. approved!

PR #1506, opened 2025-05-16
Improve Error Handling Architecture

Ezequiel-Valencia:
Introduce a flow which allows for errors to be logged, have their appropriate HTTP codes, and be included within the generated OpenAPI spec.

- [x] Introduce VCellHTTPError response object for errors
- [x] Ensure each endpoint throws the appropriate error that is expected
- [x] Utilize a factory method within the client to de-marshal the errors thrown by the server, while keeping the original function signatures
- [x] Unit tests for error handling.


Ezequiel-Valencia:
In total this PR:

- Update Quarkus server error handling architecture such that:
   - All errors thrown from handlers are web based errors, that is they inherit the base error ApiException
   - Exception mappers are used to catch all exceptions and return to the client an appropriate response
   - Logging all errors on the server that are thrown back to the client
- Utilize an error factory in the GUI client that assigns web errors to domain specific errors
- Remove git patches to fix client file receiving

Ezequiel-Valencia:
- [ ] Relocate the custom object mapper
- [ ] Ensure all original server errors are preserved when sending errors to the client, such that they can see the same message and catch the same exception 
- [ ] Remove the hard coded BioModel error used for testing.

jcschaff (approved):
This was great work.  we now have a good pattern for migrating the endpoints from RPC to REST while preserving error handling.  Approved.

PR #1520, opened 2025-05-29
Add Parameter that Prompts Login

Ezequiel-Valencia:


jcschaff (approved):
very nice solution, prompts only after logout.  perfect ... approved.

PR #1532, opened 2025-06-09
upgrading to langevin 1.4.0

danv61:
upgrading to langevin 1.4.0


jcschaff (approved):
looks good - I double-checked with vcell-fluxcd that the Langevin solver uses vcell-batch which grabs the "local" executables described within this PR.

Approved.

PR #1544, opened 2025-06-20
VC Image Resource

Ezequiel-Valencia:
The VC Image resource now uses the Quarkus API rather than the old server.


jcschaff (approved):
reviewed with @Ezequiel-Valencia, looks good - minor comments conveyed.  Approved.

PR #1550, opened 2025-06-24
VCInfo Container is Retrieved from the New API

Ezequiel-Valencia:


CodeByDrescher (approved):
Seems straightforward and consistent, but overall correctness can't be fully determined at a glance. Merge should be dependent on live-testing through VCell TEST before merge and deployment to more sensitive deployment targets.

PR #1553, opened 2025-06-27
Steady State UI changes

danv61:
Add Steady State panel in the Simulation Options dialog
Add Steady State checkbox in the Simulation table
Only for CVODE


CodeByDrescher (approved):
This gets the needed functionality out the door; will more than likely end up with future tweaks / changes, but this is more than sufficient for now!

PR #1594, opened 2025-10-02
Fix Math Model Export Error

Ezequiel-Valencia:
A type casting error was thrown when attempting to export math models.


danv61 (approved):
Looks good to me. Approved.

PR #1616, opened 2026-01-09
correct location for membrane reactions

### LINKED ISSUE #1617: correct location for membrane reactions
correct location to membrane for conditional transition reactions
when compartmental transitioning site is bound to a membrane site
PR #1616

danv61:
correct location for membrane reactions

correct location to membrane for conditional transition reactions
when compartmental transitioning site is bound to a membrane site

Fixes #1617

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -191,9 +191,9 @@
 		<solvers-vcell-mac.version>v0.0.44-dev4</solvers-vcell-mac.version>
 		<solvers-vcell-windows.version>v0.0.40</solvers-vcell-windows.version>
 		<solvers-vcell-linux.version>v0.0.44-dev4</solvers-vcell-linux.version>
-		<solvers-langevin-mac.version>1.4.5</solvers-langevin-mac.version>
-		<solvers-langevin-windows.version>1.4.5</solvers-langevin-windows.version>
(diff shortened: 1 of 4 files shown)


vcell-server/src/test/resources/slurm_fixtures/langevin/V_TEST2_999999999_0_0.slurm.sub
```diff
@@ -9,7 +9,7 @@
 #SBATCH --cpus-per-task=1
 #SBATCH --mem-per-cpu=4096M
 #SBATCH --nodes=1
-#SBATCH --time=52:49:00		# timeout for the entire job
+#SBATCH --time=6-14:24:00		# timeout for the entire job
```
jcschaff: does this mean 6 days?  we have to keep an eye on that.

danv61: that time is computed, based on "1 day" requirement for single run (as requested by Les) multiplied with the projected number of batches
ex: 100 runs, executed in batches of 20 => 5 batches of 1 day each => 5 days
i'm sure we can modify the script and let slurm reevaluate once the first few simulations are finished

PR #1625, opened 2026-01-28
Using the default singularity on batch run node

### LINKED ISSUE #1626: Using the default singularity on batch run node
Using the default singularity on batch run node 
PR #1625

danv61:
Using the default singularity on batch run node
instead of loading module explicitly

Fixes #1626


jcschaff (approved):
looks good, approved

PR #1661, opened 2026-04-14
Springsalad batch run results visualization

### LINKED ISSUE #1662: Springsalad batch run results visualization
Springsalad batch run results visualization
PR: #1661
Plots and tables
General statistics: AVG, Min/Max, SD
Cluster analysis: ACS, SD, ACO

danv61:
Springsalad batch run results visualization
Plots and tables
General statistics: AVG, Min/Max, SD
Cluster analysis: ACS, SD, ACO

Closes #1662


vcell-cli/src/main/java/org/vcell/cli/run/plotting/Results2DLinePlot.java
```diff
@@ -319,4 +327,246 @@ private static PageFormat generateAlternatePageFormat(){
         pageFormat.setOrientation(PageFormat.LANDSCAPE);
         return pageFormat;
     }
+
+
```
CodeByDrescher, line 331: Let's move this to a dedicated file in vcell-client now that jfreechart is a dependency over there. While VCell CLI does use jfreechart, having it demostrated via swing is better suited in vcell-client (especially for future programmers looking for anything to help them when developing the client); I don't think people doing client work would expect to look for a demo in CLI.

CodeByDrescher (changes requested):
Very well put together. The design paradigm introduced for future work (even beyond spring salad) seems fairly intuitive and simple.

Requesting 1 change for "posterity", but everything here looks great regardless!

CodeByDrescher (approved):
After conversation with Dan, previous changes requested can be ignored. Approved!

vcell-cli/src/main/java/org/vcell/cli/run/plotting/Results2DLinePlot.java
danv61: I evaluated JFreeChart in vcell-client but we eventually decided against using it, so the dependency is gone.

Issue #1450, opened 2025-03-04
Renaming Applications in the tree view does not implement correct "onLoseFocus" behavior


CodeByDrescher:
If the user does not confirm the rename with `enter`, the changes are lost on focus change

CodeByDrescher:
This occurs in more than 1 place actually

CodeByDrescher:
fixed with pr mentioned above

PR #1665, opened 2026-04-17
Fixing bugs when renaming in the document tree

CodeByDrescher:
Clicking off now cancels changes when renaming in the document tree


danv61 (approved):
Good catch. Approved.

CodeByDrescher:
closes #1450

PR #1667, opened 2026-04-24
Changed Spatial Results Export Selection to be CheckBox based

CodeByDrescher:
previously, selecting species to export was a delicate process. Mis-clicks when multi-selecting would result in needing to start over. On models with over 30 species, this could prove very frustrating.

The JPane now uses checkbox-like selection, and has a very rudimentary memory system.

Limitations:
- It's limited to just the session of VCell. If you quit VCell, it will reset.
- It's limited to a single run of a single simulation. Re-running the simulation will reset the selections.
- It's limited to server runs only. Local runs are considered one-off, so each blue-button simulation is a "re-run" so-to-speak.
- Switching Export Formats clears the selections.

The future export refactor should address these limitations.


danv61 (approved):
Good feature and implementation. Approved.

PR #1698, opened 2026-05-20
BioNetGen BNGL import enhancement

### LINKED ISSUE #1697: BioNetGen BNGL import enhancement
VCell does gives an error when importing BNGL models that have ~0 for states. However, such models can be imported. Can we upon import replace ~0 with ~s0 (stands for state0). Similarly, modelers with BNGL may use ~1, ~2 etc. Replace them with ~s1, ~s2 etc

danv61:
BioNetGen BNGL import enhancement
replace ~<digit>   with   ~s<digit>

Closes #1697


jcschaff (approved):
Approved, per Michael's request and your implementation

PR #1735, opened 2026-07-14
Dan ss bug fixes

danv61:
Implementing copy springsalad application
Some mesh-related issue warnings are irrelevant for springsalad application
mirroring some slurm batch langevin properties on client and server, make them required
simulation summary panel now displays meaningful (and useful) langevin simulation options
disable parameter scans for langevin simulations
optimization for RPC transfer of results for langevin batch runs
code fix when both reactants are the same molecule type
raise issue when Kon is too large


jcschaff (approved):
side-by-side review with Dan and made minor comments which have been incorporated, approved.

PR #1745, opened 2026-07-23
Upgraded ACLEditor Dialog(s)

### LINKED ISSUE #1742: File Permissions dialog box change "OK" to "Done" to avoid confusion
In the file permissions dialog box, users often add a username, then select "OK" thinking that will give view permissions to the user.  But that doesn't happen until you select "Add User".  Changing the button name to "done" may make this clearer to the user.

CodeByDrescher:
1) Fixed UX issue where it was unclear proper steps to set user access. 
2) Fixed button enable logic to assist in demonstrating when changes have been made. 
3) Fixed bug where user removed from list would not appear to be removed due to logic. 
4) Cleaned up class, added sensible names, and overall made code more readable


danv61 (approved):
UI feels friendlier.
Nice touch using names that make sense.

PR #1752, opened 2026-07-24
Rework CI into a two-tier pipeline (build-once + deferred regression)

jcschaff:
Draft for review — reworks `.github/workflows/ci_cd.yml` to cut per-push CI time and defer the heavy regression suites, per the discussion about getting the common case under ~5 min while keeping regression discipline.

## Problem

Every push to every branch ran the full 7-group matrix. Measured on a recent run:

| Job | Duration |
|---|---|
| MathGen_IT | **26 min** (40+ on a later run) |
| SEDML_VCML_IT | 17 min |
| Fast | 8.5 min |
| CI-build | 5 min |
| SBML/BSTS/Quarkus | 4–6 min |

Two compounding wastes: all 8 jobs re-ran the **full Python stack** (~2 min each) and a **full `mvn clean install` compile** (~3–4 min each); and the integration suites genuinely take 20–40 min. WIP commits paid the whole bill.

## New structure

```
build (once) ──▶ fast-tests + quarkus-tests          every push (fast lane)
build (once) ──▶ regression-tests (5-group matrix)    non-draft PR · nightly · merge queue
```

- **`build`** compiles the monorepo once, runs the Python package tests once, verifies the Docker image builds, and uploads the compiled workspace (`**/target`) as an artifact.
- **Test jobs** restore that artifact and run `mvn --offline surefire:test` — **no recompile**. Verified locally: `surefire:test -Dgroups=Fast` on `vcell-util` ran 42 tests with zero compilation.
- **Regression suites** (MathGen / SBML / SEDML_SBML / SEDML_VCML / BSTS) no longer run on ordinary pushes. Open a PR as a **draft** to stay in the fast lane; **mark it ready-for-review** to run the full regression. Nightly on master (`cron`) and a `merge_group` trigger backstop it.
- Added `concurrency: cancel-in-progress`.

## Expected effect

- Ordinary-push wall-clock: **~26–40 min → under ~10 min** (build + Fast + Quarkus).
- **Fast under 4 min** now that it doesn't recompile.
- Regression coverage preserved, moved to merge-readiness / nightly / merge queue.

## How this answers the tension

Draft PRs are cheap, so opening a PR early is encouraged; the expensive regression gate applies when you mark ready-to-merge, not on every WIP commit. The build-must-pass + Fast + Quarkus fast lane still enforces compile-clean + unit discipline on every push.

## Review notes / things to validate in a real run

- **`--offline` reuse** relies on the setup-java Maven cache carrying all external deps from the `build` job into the test jobs (they share a cache key for the same poms; `needs: build` guarantees the cache is saved first). Locally solid; worth confirming on a clean CI cache. If it's ever flaky, dropping `--offline` is a one-word change.
- **Double `build` on ready PRs:** a non-draft PR that receives a push runs `build` twice for that SHA (once via the push fast lane, once via the pull_request regression run), ~5 min each. A merge queue is the clean long-term fix and is already wired (`merge_group`). Documented inline.
- **Python in the fast lane:** Fast/Quarkus no longer set up Python. If any Fast-tagged test shells out to the Python CLI utils it would need re-adding (regression jobs keep full Python).
- **Branch-protection required checks:** the required check *names* are preserved (`CI-Test-group-Fast`, `CI-Test-group-Quarkus`, `build`). If MathGen/SBML/etc. are currently *required*, they'd need to be de-required (or moved to the merge queue) so draft/most PRs aren't blocked waiting on jobs that intentionally didn't run.
- **CodeQL** is a separate workflow (`codeql-analysis.yml`) and already only runs on PR-to-master; not touched here.
- **CD** (release) job is unchanged.


jcschaff:
Added a **manual trigger** (`workflow_dispatch`, commit 2326bf7) — a "Run workflow" button that launches the expensive regression suite on demand from any branch, with a `test-group` input to run all five groups (default) or just one. A small `prepare-regression` job computes the matrix so a single job definition serves the scheduled / merge-queue / PR / manual triggers.

Companion branch-protection change (separate, applied after this merges): the drafted master ruleset requires the full 8-check suite (evaluated in the merge queue) and blocks force pushes via `non_fast_forward` — so "full-suite-in-queue" and "disable force pushes" are both covered there, not in this workflow.

jcschaff:
**Split into three files** (commit e626b39) to fix the cluttered check list you flagged:

| file | triggers | jobs |
|---|---|---|
| `ci.yml` | push, merge_group | build, Fast, Quarkus (fast lane) |
| `regression.yml` | PR ready-for-review, merge_group, schedule, workflow_dispatch | prepare + 5 heavy IT groups |
| `cd.yml` | release | biosimulators image build/push |

Disjoint triggers mean an ordinary push now shows exactly `build` / `Fast` / `Quarkus` with **nothing skipped**; **draft PRs trigger no regression jobs at all** (the `ready_for_review` type only fires on the draft→ready transition, so the ugly skipped `CI-Test-group-${{ matrix.test-group }}` is gone); and CD never appears as a skipped check.

Check (context) names are unchanged, so the drafted master ruleset still matches. Regression jobs self-compile rather than reuse the build artifact — self-contained, and avoids a second `build` check colliding with ci.yml's in the merge queue. The merge queue runs both `ci.yml` and `regression.yml`, so the full suite still gates every merge.

Also fixed the earlier red: dropped `--offline` from the surefire test jobs (the `-DskipTests` build never fetched the surefire JUnit-platform provider, so offline resolution failed).

PR #1758, opened 2026-07-25
Fix #1746: resolve results-viewer metadata from the submit-time model

### LINKED ISSUE #1746: ODE/PDE Viewer race condition (stale results-viewer metadata → NPE in StructureAnalyzer)
## Summary

Renaming a structure (compartment/feature) to a name already used by another
structure raises the expected `ModelPropertyVetoException` ("another structure
already using name …"), but the model is then left in an inconsistent state:
the subsequent geometry/mapping refresh crashes with a `NullPointerException`
(and an `ArrayIndexOutOfBoundsException`) instead of cleanly rejecting the edit.

Reported via an auto-submitted client crash report.

## Environment

- VCell `Rel_Version_8.0.0_build_03`
- Windows 11, Java 17.0.7 (amd64)

## Steps to reproduce (inferred)

1. Open a BioModel with ≥2 structures.
2. Rename one structure to a name already in use (e.g. `Nuc`) via the Structure
   Properties panel.

## Observed

The rename is vetoed as expected, but the follow-on structure analysis throws:

```
cbit.vcell.model.ModelPropertyVetoException: another structure already using name Nuc
    at cbit.vcell.model.Model.vetoableChange(Model.java:4079)
    at cbit.vcell.model.Structure.fireVetoableChange(Structure.java:258)
    at cbit.vcell.model.Structure.setName0(Structure.java:375)
    at cbit.vcell.model.Feature.setName(Feature.java:114)
    at cbit.vcell.client.desktop.biomodel.StructurePropertiesPanel.changeName(StructurePropertiesPanel.java:518)

... followed by:
java.lang.RuntimeException: Exception caught in StructureAnalyzer.update():
  Cannot invoke "cbit.vcell.mapping.SpeciesContextMapping.isFastParticipant()" because "scm" is null
Caused by: java.lang.NullPointerException: ... "scm" is null
(also: java.lang.ArrayIndexOutOfBoundsException: Index 0 out of bounds for length 0)
```

## Expected

The duplicate-name rename should be rejected with the veto dialog only, leaving
the model unchanged and internally consistent (no NPE/AIOOBE in
`StructureAnalyzer.update()`).

## Notes

`StructureAnalyzer.update()` appears not to handle a null `SpeciesContextMapping`
after a partially-applied/rolled-back structure edit.

---
_Source: VCell support email (Message-ID `e18f3f3475b1470bb95f841acf422ba9@MN6PR14MB7148.namprd14.prod.outlook.com`). Reporter details withheld._

jcschaff:
## Problem (#1746)

The ODE/PDE results viewers build data-symbol metadata by (re)generating a math mapping for the `SimulationContext` via `getMostRecentlyCreatedMathMapping()` — a **live-editing artifact**. If the user edits the model while results are open (or between a quick-run request and the results window appearing — quick run doesn't block input and can lag), that mapping drifts from the math the result columns actually came from:

- `StructureAnalyzer` observes an inconsistent model and `getSpeciesContextMapping(sc)` returns `null` → **NPE** at `scm.isFastParticipant()` (`StructureAnalyzer.refreshTotalSpeciesContextMappings`), wrapped as `Exception caught in StructureAnalyzer.update()`.
- Even without a crash, a **deleted/renamed species** yields wrong or missing metadata, because the metadata is resolved against the edited model rather than the one that produced the results.

## Fix (Phase 1)

Bind the results viewer's `SimulationWorkspaceModelInfo` to a **pristine, submit-time `SimulationOwner`** instead of the live editable model. Because a freshly loaded/parsed copy has no cached math mapping, the existing resolver naturally calls `createNewMathMapping()` on it — so `SimulationWorkspaceModelInfo` itself is unchanged; the entire fix is in `ClientSimManager`.

New helper `resolvePristineMetadataOwner(hashTable)` resolves the owner, in order:

1. **local quick runs** — an immutable `BioModel` XML snapshot is stashed on the EDT at **run-submit** (`stashBioModelTask` in `runQuickSimulation`, right after `updateMath`, atomic w.r.t. concurrent edits), then re-parsed in `retrieveResultsTask`;
2. **saved runs** — reload a pristine `BioModel` copy from the database by version key (`documentManager.getBioModel(key)`);
3. **fallback** — unsaved model / no snapshot → the live owner (previous behavior). Never throws: any failure logs and falls back.

Resolution runs **once per dispatch** (before the per-sim loop), keyed on the shared `getSimWorkspace().getSimulationOwner()`; every viewer in a multi-select display binds to that same pristine owner (all displayed sims share one `SimulationContext`). The interactive machinery (output-function listener, `SimulationWindow`) stays on the **live** owner; only the metadata model-info uses the pristine copy.

### Why this over the first attempt
The initial commits on this branch snapshotted the *current* model at view time. That stopped the NPE but still resolved metadata against the *edited* model, so deleted/renamed species were still wrong. Snapshotting at **run-submit** (or reloading the DB copy) resolves metadata against the model the results actually came from — watertight. This PR was squashed to the single corrected commit.

## Prior manifestations (same code path)

This defect has surfaced and been band-aided before, all through `SimulationWorkspaceModelInfo.populateDataSymbolMetadata` (error `Failed to determine metadata for data symbols` → NPE):

- **#1185** (closed) — local ODE/stochastic sim results appear with the error; closed as "already fixed, see #1433".
- **#1433** (closed) — race when first displaying results after *run-before-save* / *copy-application-as-new* (Windows); patched by **#1465**, a narrow TreeMap concurrent-modification fix.

#1465 fixed one symptom but left the root cause (regenerating math from the live model) in place, so it recurred as #1746 with a different null. This PR fixes the root cause and subsumes that family.

## Follow-up (Phase 2, separate PR)
Carry simulation database keys in the cached biomodel metadata (`BioModelInfo` / child summary) so a viewed run can resolve **any** owning biomodel across versions (an incremental-save simulation leaf is non-unique). Then switch owner resolution to a sim-key ownership scan for cross-version cases.

## Testing
- `mvn compile test-compile -pl vcell-client -am` passes.
- Live-client note: the fixed viewer classes were exercised against vcell-dev (quick run + concurrent structure rename) with no NPE; a controlled old-vs-new A/B is pending on the old bytecode.


jcschaff:
## Update: cross-version metadata owner resolution (Phase 2 consumption)

Pushed `a306730d1c` — this is the follow-up that actually *uses* the Phase 2 `simKeys` (from #1761).

### What changed
`ClientSimManager.resolvePristineMetadataOwner` gains a **sim-key ownership scan** (`resolveOwnerBySimKey`): it takes the viewed simulation's authoritative key, scans the client-cached `getBioModelInfos()` for the version whose child-summary `simKeys` contain it, reloads that pristine version, and returns the `SimulationContext` that owns the key — matched **by key within the reloaded model**, so the correct application is chosen even in multi-app models.

New resolution order:
1. stashed submit-time XML snapshot (local quick runs)
2. **sim-key ownership scan (cross-version)** ← new
3. live version-key reload
4. `null` → fall back to live owner

### Why
Step 3 (the previous behavior) reloaded by the **live** model's version key. If the model was re-saved to a new version after the run was submitted, that resolved the wrong (newer) model. The scan finds the version the run actually belongs to. It's non-fatal throughout: returns `null` for unsaved quick-run sims (no key) or when no accessible model claims the key, so the remaining steps still apply, and it never throws.

### Branch caught up + dependency
This branch was 4 behind master; I merged `master` and `add_simkeys_to_infos` (#1761) in so it compiles and CI passes standalone. **This PR now contains #1761's commits**, so the intended landing order is **#1761 first, then this PR** (or this PR carries #1761 in on merge). `mvn compile -pl vcell-client -am` builds clean.

PR #1760, opened 2026-07-25
Phase 2 (#1746): add simulation keys to BioModelChildSummary via /vcInfoContainer

jcschaff:
## Why

Prerequisite consumption step for #1746 metadata correctness. To resolve a viewed simulation back to an owning BioModel *across versions* (an incremental-save simulation leaf is non-unique), the cached BioModel metadata needs the simulation database keys. Builds directly on #1759 (the `getVCInfoContainer` RPC→REST migration), which is why adding this field is now `serialVersionUID`-safe.

## What

- **Domain**: `BioModelChildSummary` gains `simKeys` (`String[]`) + `getSimKeys`/`setSimKeys`. **Read-time only** — not part of `toDatabaseSerialization`. `@JsonInclude(NON_NULL)` keeps it **absent from the other summary endpoints'** responses, so their existing (strict Python) clients are unaffected **without an opt-in flag** — the field is only ever populated on the freshly-added `GET /api/v1/vcInfoContainer` endpoint (which has no legacy clients).
- **Bulk source**: `BioModelDbDriver.getSimulationKeysForBioModels` joins `vc_biomodelsim` in batched IN-list queries (Oracle 1000-element limit), grouped by BioModel — **one bulk fetch for the whole container, never N+1**. Threaded through `DBTopLevel` and `DatabaseServerImpl`.
- **`VCInfoContainerService`** sets each biomodel's `simKeys` from that single bulk query.
- **Client mapping**: `DtoModelTransforms.dtoToBioModelChildSummary` carries `simKeys` into the domain object so the desktop can consume them.
- **Regenerated** OpenAPI spec + Java/Python/TS clients (additive: `simKeys` only, `openapi.yaml` +4 lines).
- **Test**: the API-client test asserts `vcInfoContainer` populates `simKeys` on each biomodel summary.

## No migration, no stored-format change

Keys are sourced live from the existing `vc_biomodelsim` link table, so **all historical BioModels are covered immediately** and prod/dev can deploy on independent schedules (no shared-DB coupling).

## Follow-up

The `#1746` resolver enhancement — `ClientSimManager.resolvePristineMetadataOwner` gaining a sim-key ownership scan (using these keys) as an additional resolution layer beyond the Phase 1 version-key reload — is a small follow-up on the #1758 branch, now unblocked.

## Verification
- `mvn compile test-compile -pl vcell-rest -am` and `mvn compile -pl vcell-client -am`: BUILD SUCCESS. Tests run in CI's Keycloak environment.

### DIFF
--- a/python-restclient/.openapi-generator/FILES
+++ b/python-restclient/.openapi-generator/FILES
@@ -119,8 +119,6 @@ docs/VcelloptStatus.md
 docs/Version.md
 docs/VersionFlag.md
 test/__init__.py
-test/test_vc_info_container_resource_api.py
-test/test_vc_info_container_summary.py
 tox.ini
 vcell_client/__init__.py
 vcell_client/api/__init__.py
--- a/python-restclient/docs/BioModelChildSummary.md
+++ b/python-restclient/docs/BioModelChildSummary.md
@@ -11,6 +11,7 @@ Name | Type | Description | Notes
 **app_types** | [**List[MathType]**](MathType.md) |  | [optional] 
 **sim_names** | **List[List[str]]** |  | [optional] 
 **sim_annots** | **List[List[str]]** |  | [optional] 
+**sim_keys** | **List[str]** |  | [optional] 
 **geometry_dimensions** | **List[int]** |  | [optional] 
 **geometry_names** | **List[str]** |  | [optional] 
 **simulation_context_annotations** | **List[str]** |  | [optional] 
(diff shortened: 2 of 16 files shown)


jcschaff:
Superseded by #1761. That branch (`phase2-biomodel-simkeys`) used the earlier holder + `String[][]` design; #1761 is the reworked flat-`String[]`, single-query implementation on `add_simkeys_to_infos`.

PR #1763, opened 2026-07-26
windowing: owned modeless child windows (LWChildDialog) — fixes child-window z-order

jcschaff:
## Owned modeless child windows — systemic z-order fix

Implements `docs/windowing-design-patterns.md` §7.4 (merged in #1762): move modeless child windows off the fragile **un-owned `JFrame` + `Window.toFront()`** path onto a **native owned-window** relationship the OS enforces.

### Why
A `JFrame` cannot have an owner in AWT, so `LWChildFrame` could only be kept above its parent by a best-effort `toFront()` — which modern **macOS** (13.3+/Sonoma cooperative activation) and **Windows** (foreground lock) refuse for a non-foreground app. An owned window is kept above its owner **by the OS itself**. (Full platform analysis + JDK bug IDs in §7.)

### What
1. **`LWChildDialog`** (new) — the owned counterpart to `LWChildFrame`: an owned modeless `JDialog` (parent's window as native owner + `MODELESS`), implementing `LWContainerHandle`/`LWFrameOrDialog` and registering with `LWManager` identically. No `toFront()` dependence.
2. **`ChildWindowManager.ModelessChild` → `LWChildDialog`** — the systemic fix. Every `MODELESS` child created via `addChildWindow(...)` is now owned. This is the high-leverage change: it covers the general child-window path, not a single window.
3. **`AddAnnotationsPanel`** — a specific P1 audit violation (raw un-owned `JFrame` with a tell-tale commented-out `setAlwaysOnTop`) converted to extend `LWChildDialog` as a concrete example.

### Why it's safe
- `ChildWindow` drives `impl` purely through the `LWFrameOrDialog` interface — no `Frame` cast.
- The **modal** sibling `ParentModalChild` already extends an owned `JDialog` (`LWTitledDialog`) and works in this exact machinery — this just extends the proven pattern to the modeless case.
- Close/re-open is handled by the `windowClosing` listener (`closeChildWindow` → dispose + remove), independent of the default close operation.
- `vcell-client` compiles (BUILD SUCCESS).

### Trade-off (ratified by maintainer)
Owned windows have **no independent taskbar button** and travel with their owner under Spaces / Stage Manager / Snap / virtual desktops. For child editors/viewers that logically belong to a document window, that is the desired behavior.

### Verification note
Runtime z-order can only be confirmed interactively on macOS/Windows (the initiative's headless-observation crux). Compilation + the structural argument above are what CI can check.

### Follow-ups (not in this PR)
- `VcellHelpViewer` (P1) is a *global* help window, a poor fit for owned-child — it should become an `LWTopFrame` (LW-tracked top-level) instead. Separate change.
- The six orphan-`JDialog`+`alwaysOnTop` dialogs and null-parent `JOptionPane` sites (§6 P1/P2) are `DialogUtils`-routing fixes, tracked separately.

### DIFF
--- a/vcell-client/src/main/java/cbit/vcell/client/ChildWindowManager.java
+++ b/vcell-client/src/main/java/cbit/vcell/client/ChildWindowManager.java
@@ -19,6 +19,7 @@ import javax.swing.JMenuBar;
 
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
+import org.vcell.client.logicalwindow.LWChildDialog;
 import org.vcell.client.logicalwindow.LWChildFrame;
 import org.vcell.client.logicalwindow.LWContainerHandle;
 import org.vcell.client.logicalwindow.LWFrameOrDialog;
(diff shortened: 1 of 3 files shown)


jcschaff:
**Interactive validation (macOS).** Confirmed the core mechanism with a standalone, server-free harness: one `LWTopFrame` parent opening an old un-owned `LWChildFrame` child vs a new owned `LWChildDialog` child.

- **OLD** `LWChildFrame` (today's `ModelessChild`): sinks **behind** the parent when the parent is clicked / after switching apps and back — the reported z-order bug.
- **NEW** `LWChildDialog` (what this PR makes `ModelessChild` extend): stays **in front**, OS-enforced. The accepted trade-off (no independent Dock/taskbar entry) is also visible.

This validates the `LWChildDialog` primitive's z-order behavior directly. Note: the systemic `ChildWindowManager.ModelessChild` integration could **not** be exercised in the full desktop client during this session because the from-source client (built off current master with the #1759/#1761 REST changes) is ahead of the `vcell-dev` server deployment → unrelated API version-skew. Full in-app confirmation should follow once the #1746 REST arc is deployed to dev.

PR #1791, opened 2026-07-28
ci: speed up CodeQL Java ~17% (manual parallel build, tests skipped)

jcschaff:
**Draft to measure — do not merge until the timing is in.**

## Background (phase profiling of the autobuild baseline)

| Phase | autobuild (baseline) |
|---|---|
| build + extraction | **6m 30s** ← bottleneck |
| query / analysis | 3m 02s (roughly fixed) |
| **total** | ~9m 24s |

The query phase is *not* the bottleneck — the serial `javac` over the whole monorepo + dependency resolution is. (Also learned: the reverted `build-mode: none` still ran Maven; its slowness came largely from a dropped maven cache → ~1268 Central re-downloads.)

## What this experiment changes

- **Keep the warm maven cache**, guarded to Java (JS/Python don't need a JDK).
- **`build-mode: manual`** for Java (`none` for JS/Python).
- Replace `Autobuild` with an explicit **parallel, test-skipping** compile traced by CodeQL:
  `mvn -B -T1C -DskipTests -Dmaven.test.skip=true -Dmaven.javadoc.skip=true install`
  - `-T1C` parallelizes across cores.
  - skipping test compilation cuts both compile time *and* extraction of test sources.

## What we're measuring

Does the Java `Analyze` job beat **9m 24s**, and by how much? Compare the build+extraction phase specifically.

## Trade-off

Test code is no longer scanned by CodeQL (low value). If we want it back, drop `-Dmaven.test.skip=true` (keeps test compile, still parallel).

## Further levers (not in this draft, to keep the measurement clean)

- `paths-ignore` the generated `vcell-restclient` client (reduces alert noise; minor speed effect on traced builds).

CI-config only.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -40,7 +40,12 @@ jobs:
     - name: Checkout repository
       uses: actions/checkout@v4
 
+    # Only Java needs a JDK + warm maven cache. The maven cache is essential:
+    # even CodeQL's own build (autobuild OR the manual build below) resolves the
+    # full dependency classpath via Maven, and a cold cache re-downloads ~1200+
+    # artifacts from Central (measured), which dominated the extraction phase.
     - name: setup java 17 with maven cache
+      if: ${{ matrix.language == 'java' }}
       uses: actions/setup-java@v4
       with:
         distribution: 'temurin'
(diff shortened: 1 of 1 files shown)


jcschaff:
## Measured result — adopting

| Phase | autobuild (baseline) | manual parallel (this PR) | Δ |
|---|---|---|---|
| build + extraction | 6m 30s | **5m 10s** | −1m 20s |
| query / analysis | 3m 02s | 2m 09s | −0m 53s |
| Maven Central downloads | 182 | **1** (warm cache) | — |
| **total** | **9m 24s** | **7m 47s** | **−1m 37s (~17%)** |

Job passed; the saving lands in the build phase exactly as predicted (`-T1C` parallel compile + skipped test compilation). Test code is no longer CodeQL-scanned (accepted trade-off — low security value; main + generated code still fully covered).

PR #1813, opened 2026-08-01
SaLaD 3D viewer phase 1.5: species selector, MP4 export, color fix

jcschaff:
Refinements to the SpringSaLaD 3D trajectory viewer, plus a verified build guide.
Seven commits, each self-contained.

## Reported bug: some species render gray

`SpringSaladViewerCanvas` carried its own hand-written 14-entry color table with a
light-gray fallback, while VCell writes the solver input from the canonical 28-entry
`org.vcell.util.springsalad.Colors`. Everything the smaller table missed rendered
gray — not just `LIME` and `LIME_GREEN` as reported, but `PURPLE`, `MAROON`, `NAVY`,
`OLIVE`, `TEAL`, `GOLD`, `DARK_GREEN`, `CRIMSON`, `DARK_VIOLET`, `VIOLET`,
`SLATE_BLUE`, `LIGHT_CYAN` and `DARK_CYAN`. About half the palette.

Colors now resolve from `Colors` directly so the two tables cannot drift apart again.
`BLACK` is the one deliberate deviation: the sphere sprite multiplies color by
luminance, so pure black would be an invisible disc on the black background.

## Oblique default camera

The camera looked straight down the z axis, which shows the membrane face-on and,
under a parallel projection with no perspective cue, flattens the box. It now starts
35° above the membrane and 30° around it.

Two latent bugs surfaced doing this:

- `resetView()` only reset zoom and pan. Rotation is driven by the trackball's
  quaternion, not the camera, so `camera.resetView()` never restored orientation.
- Framing scaled by the largest single axis extent, so a rotated box ran its diagonal
  off the canvas. It now frames on the bounding sphere, which fits at any rotation.

Worth knowing: `Trackball.setRotation` composes its Euler angles as `Rz*Ry*Rx`, so the
turntable azimuth is the **Y** angle — the Z angle is a roll that tilts the horizon.

## Species selector

A checklist beside the view for showing and hiding site types, each decorated with the
same shaded ball the renderer draws for it. Hidden sites are dropped before projection
so bonds to them are skipped rather than left dangling, and they still count toward the
scene bounds so framing does not jump as types are toggled.

Naming them took some digging. The trajectory file identifies a site only by id, radius
and color, so the first implementation grouped by appearance — which merges two
molecules whose sites share a color and radius. That is not hypothetical: in the
solver's own `all_reactions` fixture, `MT0` and `MT1` both have a `DARK_GRAY` r=1 site
and a `BLUE` r=1 site.

The solver already writes what is needed: `MySystem.writeSiteIDs()` emits a
`SiteIDs.csv` naming every site it created, including molecules created part-way
through a run by creation reactions. The data service now reads it alongside the viewer
file and returns an augmented `SpringSaladTrajectory`.

Deliberately optional at every step, so old datasets keep working:

- runs from before this file was served, and runs whose solver folder has been pruned,
  come back with no identities and the viewer falls back to grouping by color and
  radius, with names joined from the `MathDescription` where they can be
- a malformed legend is logged and skipped rather than costing the user the trajectory
- `SpringSaladTrajectory` keeps `serialVersionUID` 1L and only gains a field, so it
  stays compatible over JMS `ObjectMessage` in both directions

No new RPC method, so no new delegation in the four `DataSetController` implementations,
and no solver change — which is what keeps the standalone SpringSaLaD application that
shares the solver unaffected.

The alternative considered was reconstructing identity from the site id, which the
solver composes as `moleculeInstance * 10000 + siteIndex`. It does work — verified
against the fixture, 248/248 sites — but it bakes in the solver's internal id scheme,
its molecule creation order, and an assumption that only one species has a creation
reaction, and it would mislabel **silently** if any of those changed.

Companion solver PR, which makes archived runs keep their names once the `_FOLDER` is
pruned: cam-center/LangevinNoVis01#40. Nothing here depends on it — the `_FOLDER`
fallback covers local and un-pruned runs today.

## MP4 export

A "Save movie…" button writes the whole trajectory in the view currently on screen —
same rotation, zoom, species selection and scene toggles — at the transport bar's
playback rate.

**This adds a dependency:** `org.jcodec:jcodec-javase:0.2.5`, a pure-Java H.264 encoder
(BSD), so nothing native has to be installed. VCell's only existing movie writer is the
legacy gloworm QuickTime one, which is neither MP4 nor reusable here. Animated GIF comes
along through the in-tree `GIFUtils`. **Happy to drop this commit if the dependency is
unwelcome** — the rest of the PR is independent of it.

Frame size rounds down to whole 16-pixel macroblocks: JCodec pads to macroblocks, so a
partial one leaves the decoded frame larger than the container advertises (measured 192
against a requested 201) and players that ignore the H.264 crop show padding. Verified
by decoding the output back in a test and independently with `ffprobe` (h264, 240x176,
10 fps, 120 frames).

Rendering runs off the EDT under the standard modal progress dialog. Making that safe
meant splitting `paintScene(g, w, h, frame)` out of `paintComponent`, so exporting no
longer resizes the live component or changes the displayed frame.

The scene toggles moved from the transport bar into the sidebar next to the species
list — it groups every "what is drawn" control in one place, and the bar no longer
overflows and silently drops the time readout in a narrow window.

## Build documentation

Separate concern, included because a fresh worktree made it painful.

The build instructions were spread across four documents that disagreed with each other
and with CI: `README.md` said Maven first and listed four Python packages,
`README_BioSim.md` said Python first, `README_stack_on_mac.md` listed seven and put
Maven first, and none mentioned `pip install -r requirements.txt`. Two omitted
`dependency:copy-dependencies`, which is required rather than optional — it populates
`target/maven-jars/`, which `vcell.sh` and the root `Dockerfile` put on the classpath,
so leaving it out compiles fine and fails at run time.

`docs/BUILDING.md` mirrors `.github/workflows/ci.yml` and says so, so there is one place
to check when they drift. Every command in it was run in a fresh worktree.

It records the trap that cost the most time here: **Java tests failing on a missing
Python module or solver binary are the skipped Python step, not a broken environment.**
`vcell-core`'s Fast group reported 10 errors across `MathOverrideRoundTripTest`,
`CopasiOptimizationSolverTest` and `VCellDataTest`; after building the Poetry packages
all 433 passed with no other change, because Fast tests shell out to those environments
at run time.

## Testing

24 new tests across `vcell-core` and `vcell-client`; `vcell-core`'s full Fast group is
433 passing. SpringSaLaD test fixtures are trimmed from a real `all_reactions` solver
run, keeping the awkward parts: the color-identical `MT0`/`MT1` and a molecule created
mid-run.


jcschaff:
Superseded by #1814 — same commits, branch renamed from `worktree-sprintsalad-vis` to `feature/salad-viewer-phase1.5`. GitHub's branch rename closed this PR rather than retargeting it.

PR #1842, opened 2026-08-06
Open a producer session's JMS connection on first use, not per message

jcschaff:
Follow-up to #1837 / #1838: removes the structural cost that turned one bad export
progress value into a connection storm.

## The problem

`ConsumerContextJms` builds a `MessageProducerSessionJms` for **every message it
receives** and hands it to the listener. The constructor opened a JMS connection, a
session and a temporary queue. Most listeners never send through it:

| consumer | container | uses the session? |
|---|---|---|
| `ClientTopicMessageCollector` (ClientStatusTopic) | `api` | no — parameter never referenced |
| `SimDataServer` (pooled) | `data` | no — `VCPooledQueueConsumer` substitutes its own |
| `DatabaseServer` (pooled) | `db` | no — likewise |
| `HtcSimulationWorker` (pooled) | `submit` | no — likewise |
| `JavaSimulationExecutable` (control topic) | solver | no |
| `SimulationDispatcher` worker events + sim requests | `sched` | **yes** |

Five of seven paid a connection, a session and a temporary queue per message for
something they never touched. During the 2026-08-06 incident that reached ~3,300
connections a minute against **zero** transport interruptions.

## The change

- The connection and session open on **first use** instead of in the constructor.
- The temporary queue is created only when `sendRpcMessage` needs a reply destination
  — nothing else uses one.
- `commit()` / `rollback()` / `close()` on a session that was never used are no-ops.
- `createProducerSession()` calls `open()` explicitly: those sessions are long-lived and
  always used, so they still open at startup and a broker problem surfaces there rather
  than at the first send.

Senders behave exactly as before — same session, same transaction boundaries.

## Verification

`MessageProducerSessionJmsTest` runs a real embedded broker and counts every connection
opened. Controlled by reverting the production files to `master` with the test kept:

```
producerSessionOpensNoConnectionUntilItIsUsed  expected: <0>  but was: <1>
anUnusedProducerSessionCostsNothing            expected: <2>  but was: <3>
consumerOpensNoConnectionPerMessage            expected: <5>  but was: <10>
```

Five messages, five extra connections — the per-message cost, quantified. The fourth
test, `createProducerSessionOpensEagerly`, passes both ways: it pins the behaviour this
change deliberately preserves.

Full `vcell-server` Fast group: 56 passed.

## Not addressed here

- `sendRpcMessage` still builds a second `MessageProducerSessionJms` purely to create the
  message (its own connection per RPC). The session-sharing constructor for that is
  commented out in the source with no explanation of why, so changing it needs more than a
  guess. It no longer creates a temporary queue, at least.
- `sendTopicMessage` creates a `MessageProducer` and never closes it. Harmless while
  sessions are short-lived; it would matter if these were ever pooled.
- `removeMessageConsumer` stops the polling thread but never closes the consumer's
  connection.

https://claude.ai/code/session_01SY1XHgTZXZPqUECo2VBAWg

### DIFF
--- a/vcell-server/src/main/java/cbit/vcell/message/jms/MessageProducerSessionJms.java
+++ b/vcell-server/src/main/java/cbit/vcell/message/jms/MessageProducerSessionJms.java
@@ -38,7 +38,6 @@ import cbit.vcell.mongodb.VCMongoDbDriver;
 import cbit.vcell.resource.PropertyLoader;
 
 public class MessageProducerSessionJms implements VCMessageSession {
-    //		private static int tmpQCnt = 0;
     private VCMessagingServiceJms vcMessagingServiceJms = null;
     private TemporaryQueue commonTemporaryQueue = null;
     private Connection connection = null;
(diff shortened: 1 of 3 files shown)


jcschaff:
## Follow-up: does deferral add complexity?

Raised in review — worth answering precisely, because *sharing* and *deferring* are
different trades.

**No sharing was introduced.** Ownership is still strictly 1:1: each session opens its own
connection, closes it, and never hands it to anyone. The complexity the original
per-message design was avoiding — shared connections, coupled lifetimes, cross-thread
session reuse — is untouched.

**Deferral is not free, though.** Three costs, and where each lands:

1. *Two-state object.* Fields are written after construction rather than in it, so
   `getSession`/`getReplyQueue`/`commit`/`rollback`/`close` are now `synchronized`.
2. *`getSession()` holds the monitor while connecting*, which a `failover:` URL can make
   slow. This only bites if a **shared** session materialises lazily under concurrency,
   which is why `createProducerSession()` opens eagerly. Verified: all four shared sessions
   (`SimDataServer`, `DatabaseServer`, `HtcSimulationWorker`, `SimulationDispatcher`) come
   from `createProducerSession()`; the only other construction sites are `ConsumerContextJms`,
   which are single-threaded.
3. *Failure timing moved* — and this one was a real defect, now fixed in b570a8e.

### The defect deferral introduced (fixed)

Before: broker trouble → constructor throws → `ConsumerContextJms` catches it **before the
listener runs** → no commit → message redelivered.

After deferral: the listener runs, calls `sendQueueMessage`, `getSession()` throws — and
`sendQueueMessage` catches `JMSException` and returns normally. The listener is told the
send succeeded and the inbound message gets committed. A new way to drop a message.

Opening is now done outside that catch and reported as `VCMessagingException`, which both
send methods already declare. A failure of `send()` itself is still swallowed exactly as
before; only the open is escalated.

Pinned by `aSendThatCannotOpenAConnectionThrows`. Control against the previous commit:

```
Expected cbit.vcell.message.VCMessagingException to be thrown, but nothing was thrown.
```

Only that test fails without the fix, so the change is targeted. `vcell-server` Fast group:
57 passed.

Issue #143, opened 2022-04-21
HDF5 export settings dialog


ACowan0105:
Remove text at bottom of dialog box for HDF5 settings "See VCell Help (Biomodel.....)

danv61:
Replaced the message text as directed, the tooltip stays the same.

Issue #142, opened 2022-04-19
Help for Spatial HDF5 export


ACowan0105:
Need to update Help for HDF5 export and import using FIJI/ImageJ

ACowan0105:
Done in 8e10ea33b0bcd7d00303212c5b715792850c8a96

PR #1845, opened 2026-08-06
Give each RPC its own JMS session so concurrent callers cannot corrupt each other

jcschaff:
Fixes the concurrency defect surfaced by the forensics in #1844.

## The defect

`RpcService` creates **one** producer session at startup (`VCellApiMain:233`) and
`RpcRestlet` hands that same instance to every request thread. `sendRpcMessage` then ran
`createProducer`, `send`, `commit`, `createConsumer` and a blocking `receive` on that one
`javax.jms.Session` — which is not thread-safe. The sharpest edge is the transaction: the
session is transacted, so **one caller's `commit()` commits another caller's in-flight
send**.

Running this PR's test class against the previous implementation, with six callers at once:

```
   5  Transaction 'TX:<id>' has not been started
   9  Server is temporarily not responding      (RPC timeouts)
   → concurrentRpcsEachGetTheirOwnAnswer: ERROR
```

Marking the method `synchronized` is not the answer — it would serialise every RPC behind
a `receive()` that blocks for up to the client timeout. That is presumably why the
`synchronized` on the signature was commented out rather than kept.

## The change

`Connection` is thread-safe; `Session` is not. Each call now takes **its own session for
the whole RPC** — request message, producer, commit, reply consumer — and closes it in the
`finally`. No new connection: a session is created on the connection already held.

This subsumes the separate message session from #1844. With a per-call session there is
nothing left to isolate message construction *from*, so that indirection collapses into
this one.

The reply queue stays shared per producer session (each caller takes only its own reply,
selected by correlation id) and is now created on a session of its own, so `sendRpcMessage`
no longer touches the shared session at all. A temporary destination belongs to the
connection, not to the session that created it.

Also fixes a latent NPE — `replyConsumer` was closed in a `finally` with no null check, so
a failure in `createConsumer` masked itself.

## Verification

Two new tests, both round-tripping through a real embedded broker with a
`VCRpcMessageHandler` responder:

- `rpcRoundTripReturnsTheAnswer` — request out, reply back through the temporary queue.
  Proves the reply path still works after moving where the temporary queue is created.
- `concurrentRpcsEachGetTheirOwnAnswer` — six simultaneous callers on one producer session;
  each must get back its own payload. Passes here; on `master` it errors with the
  transaction corruption above.

`vcell-server` Fast group: 60 passed. `vcell-api` compiles.

## Two things found while writing the tests

- **`removeMessageConsumer` leaves a consumer attached.** It stops the polling thread but
  never closes the consumer or its connection, so the broker keeps dispatching to it —
  with `prefetchLimit=1` it sits on one message forever. This cost me a test failure that
  looked like a concurrency bug and wasn't: a responder torn down by an earlier test was
  still holding a prefetched message. The tests now use a queue per test to route around
  it. **This is a real production hazard, not just a test artifact**, and is still open.
- The test harness needs `setTrustAllPackages(true)`: `VCRpcRequest` travels as an
  ObjectMessage and ActiveMQ refuses to deserialize unlisted classes.

https://claude.ai/code/session_01SY1XHgTZXZPqUECo2VBAWg

### DIFF
--- a/vcell-server/src/main/java/cbit/vcell/message/jms/MessageProducerSessionJms.java
+++ b/vcell-server/src/main/java/cbit/vcell/message/jms/MessageProducerSessionJms.java
@@ -135,110 +135,112 @@ public class MessageProducerSessionJms implements VCMessageSession {
     }
 
     /**
-     * The temporary queue is a reply destination for sendRpcMessage and nothing else, so it
-     * is created with the first RPC rather than alongside the session.
+     * The temporary queue is a reply destination for sendRpcMessage and nothing else, so it is
+     * created with the first RPC rather than alongside the session.
+     *
+     * It is created on a session of its own so that sendRpcMessage never touches the shared
+     * session: a temporary destination belongs to the connection and outlives the session that
+     * created it (ActiveMQ's createTemporaryQueue delegates to the connection).
      */
     private synchronized TemporaryQueue getReplyQueue() throws JMSException{
         if(commonTemporaryQueue == null){
-            commonTemporaryQueue = getSession().createTemporaryQueue();
+            Session replyQueueSession = getConnection().createSession(false, Session.AUTO_ACKNOWLEDGE);
+            try {
+                commonTemporaryQueue = replyQueueSession.createTemporaryQueue();
+            } finally {
+                replyQueueSession.close();
+            }
         }
         return commonTemporaryQueue;
     }
 
-    public /*synchronized*/ Object sendRpcMessage(VCellQueue queue, VCRpcRequest vcRpcRequest, boolean returnRequired, long timeoutMS, String[] specialProperties, Object[] specialValues, UserLoginInfo userLoginInfo) throws VCMessagingException, VCMessagingInvocationTargetException{
+    /**
+     * Everything one RPC does -- building the request, the producer, the commit, and the
+     * consumer it then blocks on -- happens on a JMS session created for that call alone.
+     *
+     * RpcService creates a single producer session at startup and RpcRestlet hands it to every
+     * request thread, so doing this work on the shared session meant concurrent createProducer,
+     * send, commit, createConsumer and receive on a javax.jms.Session, which is not thread-safe;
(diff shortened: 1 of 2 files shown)


jcschaff:
## Note on the failing CodeQL check — pre-existing, not introduced here

CodeQL reports "1 new alert including 1 critical severity security vulnerability". It is
**alert #142, `java/unsafe-deserialization`, created 2023-11-26 and already open on
`refs/heads/master`.**

The sink is `((ObjectMessage)jmsMessage).getObject()` at `VCMessageJms.java:177` — a file
this PR does not touch. The alert's dataflow runs

```
replyConsumer.receive()  ->  new VCMessageJms(replyMessage, ...)  ->  getObjectContent()  ->  getObject()
```

and this PR moved those three lines: restructuring `sendRpcMessage` unindented them out of
the nested `else` block. They are semantically identical before and after — same receive,
same wrapper, same accessor. GitHub attributes the pre-existing alert to the PR because the
path now passes through changed lines.

Verified rather than assumed:
- `gh api .../code-scanning/alerts/142` → `created_at: 2023-11-26`, open on `refs/heads/master`
- the repo has 2 open alerts of this rule, both in `VCMessageJms` (lines 177 and 422)
- `git diff origin/master -- MessageProducerSessionJms.java` shows the three lines as
  move-only changes

An earlier revision of this branch did add a *new* instance of the pattern — the test
factory called `setTrustAllPackages(true)`, mirroring production. That was narrowed in
7c421a8 to just the packages the test serialises, so no new call site is introduced by
this PR.

CodeQL is not among this repo's required checks (`build`, `CI-Test-group-Fast`,
`CI-Test-group-Quarkus`, `regression-gate`). Merging with that understanding.

**Still worth fixing separately:** production calls `setTrustAllPackages(true)` in
`VCMessagingServiceActiveMQ:30` and `OptimizationBatchServer:45`, which is what alerts #142
and #143 are really about. Narrowing that to an allowlist is a genuine security improvement,
but VCell puts arbitrary domain objects on JMS, so it needs care.

Issue #1859, opened 2026-08-07
3D field viewer: nice-to-haves (labels, color bar, axes, cut plane, picking, time plots, movie export)


jcschaff:
Follow-ups requested after the viewer was confirmed working in an installed client (`-Dvcell.fieldViewer.enabled=true` in `vmoptions.txt`). Shipped in #1829 / #1854.

Grouped by what they actually need, because the effort is very unevenly distributed.

## A. Small, client-side only

- [x] **(1) Show the simulation name and identifier**, in the window title and/or on the page. The field server's registry is keyed `simKey:jobIndex` and `/info` does not currently carry a human-readable name — `PDEDataViewer` knows it at registration time, so this is a small addition on both sides.
- [x] **(7) Smaller minimum canvas** so several viewer windows tile on one screen. Pure CSS: `.canvas-box` is currently `max-width: 960px` with a fixed `4/3` aspect ratio.

## B. Annotation actors — one probe decides these

- [x] **(2) Color bar**
- [x] **(3) Bounding box with ticks**

`vtkScalarBarActor` and `vtkCubeAxesActor` **are compiled into the bundle** — it is an all-modules build (`VTK_BUILD_ALL_MODULES` via VTK's own `configure_wasm32_emscripten_linux.cmake`), not the trimmed set once planned. Confirmed by symbol search.

**But compiled-in is not the same as constructible.** The standalone session builds objects through the serialization registry, and we have already been bitten by that distinction (`vtkDoubleArray.setValue` is not in the invoker whitelist; `setTuple1` is). **First step for both: a ~10-minute browser probe that tries to construct each actor and calls one setter.**

Fallbacks if the probe fails:
- Color bar is genuinely *better* in HTML — crisper text, themeable, no wasm round-trip — since it is a static legend, not 3D geometry.
- The box has a fallback (`vtkOutlineFilter`, also present), but **ticks do not**: labels that track a rotating 3D scene are exactly what `vtkCubeAxesActor` exists for, and faking them in HTML means reimplementing projection. If the probe fails, this one is worth a bundle rebuild — with the caveat that owning that build was a multi-day toolchain fight (see `docs/3d-renderer-design.md`).

## C. Pipeline work, classes confirmed present

[...]

PR #1860, opened 2026-08-07
3D field viewer: label each window with its run, let the canvas tile (#1859 item A)

jcschaff:
Implements item A of #1859 — the two small client-side follow-ups from the first installed-client test of the field viewer.

## (1) Simulation name and identifier

- `PDEDataViewer` builds a human-readable `model::app::sim` name (via `SimulationModelInfo`, falling back to the simulation's own name) and attaches it when registering the dataset with `FieldViewerServer` — registration is the only moment anyone holds a name; the data files carry only keys.
- `/info` now serves `simName` (when present) and `jobIndex` alongside `simId`.
- The viewer shows the name as a header above the canvas with the `SimID_…` identifier dimmed beside it, and mirrors it into the browser-tab title; `· job N` is appended only for parameter-scan jobs (N > 0). Rendered with `textContent` — sim names are user-authored.

## (7) Smaller canvas so several viewer windows tile

The culprit was the fixed 4/3 aspect ratio, not the 960px max-width: a viewer tiled into a half-screen window got a canvas tall enough to push its own controls below the fold. The canvas box is now clamped to the viewport height (`max-height: calc(100dvh - 170px)`, 160px floor), and a debounced window-resize handler re-syncs the vtk drawing buffer to the box — the clamp changes the box's aspect ratio, and a buffer left at the old shape would draw stretched. (vtk's own resize observer stays deliberately disabled; this is the viewer's hand-rolled handler, like the camera interaction.)

Also carries the vtk.wasm capability-probe notes in `webapp-viewer/README.md` (which classes the standalone session can actually construct — groundwork for item B).

## Testing

- `vcell-client` compiles clean; `viewer.js` passes `node --check`.
- `PDEDataViewer.java` is CRLF — diff audited, zero line-ending churn.
- Not exercised in a running client (needs the wasm bundle + a finished local sim with `-Dvcell.fieldViewer.enabled=true`); the feature remains off by default.

Closes nothing — #1859 stays open for items B–F; tick items (1) and (7) there on merge.

https://claude.ai/code/session_01FYqrC3BiB4JRsJoz7BXJF4

### DIFF
--- a/vcell-client/src/main/java/cbit/vcell/client/data/PDEDataViewer.java
+++ b/vcell-client/src/main/java/cbit/vcell/client/data/PDEDataViewer.java
@@ -1708,7 +1708,8 @@ private void openInBrowserViewer() {
 		// register the dataset before opening the browser, and hand over the reader this window already
 		// uses -- that is what lets one server serve several windows, local and remote alike
 		FieldViewerServer.register(simDataId,
-				((ClientPDEDataContext) dataContext).getDataManager().getVCDataManager(), mathDescription);
+				((ClientPDEDataContext) dataContext).getDataManager().getVCDataManager(), mathDescription,
+				displayNameForViewer());
 		int port = FieldViewerServer.start();
 		if (port < 0) {
 			DialogUtils.showErrorDialog(this, "Could not start the local field viewer server. See the log for details.");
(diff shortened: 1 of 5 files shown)


jcschaff:
Demo screenshots — real renders through this branch's viewer code (vtk.wasm pipeline included); only the field server was mocked with a synthetic ellipsoid dataset, since the real one needs a finished local sim.

**Full window (1100×900)** — the new run-title header (name bold, `SimID` dimmed beside it); the browser tab title carries the same name. The height clamp is already visible here: the canvas is wider than 4/3 because the window height caps it, undistorted thanks to the buffer re-sync.

full window

**Tiled half-screen window (620×480)** — without the clamp the canvas alone would be ~465px tall and push every control below the fold; with it the canvas caps at ~310px and the controls stay on screen. (At this extreme width the control rows wrap; a realistic ~800px tile doesn't.)

tiled window

PR #1873, opened 2026-08-09
SpringSaLaD math comparison, and a math/biology/solver layer boundary

jcschaff:
SpringSaLaD math descriptions could not be compared. All three Langevin `compareEqual` overrides were stubs:

```java
LangevinParticleMolecularComponent other = (LangevinParticleMolecularComponent)obj;
if(false) {			// TODO: compare everything that needs comparing
    return false;
}
return super.compareEqual(obj);
```

`if(false)` is dead and `other` is never read, so every SpringSaLaD-specific field was invisible to comparison. Two maths differing in a site's radius, diffusion rate, location or coordinate, or a reaction's bond length, subtype or transition condition, reported as **identical**.

That is the dangerous direction. `compareEqual` decides whether a math is saved, and `testEquivalency` decides whether `SimulationVersion.parentSimulationReference` survives — so a false "identical" silently drops an edit, and a false "equivalent" leaves results computed with the *old* values attached to the new math.

Fixing it properly needed the layer boundary sorted out first, so the PR does both.

---

## 1. `docs/architecture-layers.md` — the rules, written down

The layers model the same domain with near-identical vocabulary, and several long-standing violations read as precedent, so the rules aren't recoverable from the code. They are now written down, with the dependency spine `biological → mathematical → solver`:

- **P1** a math description stands alone — *corollary: everything the solver reads must be persisted in it*
- **P2** duplicate shared concepts into math under a distinct name (`Particle*`), never import — extends to enums
- **P3** every scalar is an `Expression`; `Double → Expression → double` across the boundaries is correct
- **P4** solver input formats are not math's concern
- **P5** translate only at the two boundaries, and fail loudly at the solver one
- **P6** three comparison outcomes with three different user-visible consequences

Each rule is marked **[stated]** or **[inferred]** so the inferred ones can be corrected rather than ossifying, and known violations are listed rather than hidden.

## 2. The boundary

**18 biological imports in `cbit.vcell.math` → 0.**

The codebase already had the answer: NFSim duplicates every rule-based biological entity into math under a `Particle*` prefix, **including enums** — `ParticleBondType` mirrors `BondType` constant for constant. SpringSaLaD was the one place that wasn't carried through, so `LangevinParticleJumpProcess` now owns `ParticleSubtype` and `ParticleTransitionCondition`, with `LangevinMathMapping.toMath(...)` translating on the **VCML wire name** so a reorder fails loudly instead of silently mis-mapping.

Also moved: naming constants into `MathNamingConventions`; `MathDescription`'s 37 validation messages into `MathDescriptionMessages`; `updateReservedSymbols` now takes `Map<String,Expression>` rather than `Model.ReservedSymbol[]`.

**P4:** `LangevinParticleMolecularComponent.writeType` / `.writeSite` emitted `.lngv` syntax from inside math and were the only reason it imported `org.vcell.solver.*`. Both moved to `LangevinLngvWriter`.

`MathNamespaceSeparationTest` enforces both directions and fails when an accepted exception goes **stale**, so the list can't drift into fiction. Verified to fail on injected biological *and* solver imports.

**The shared `RuleAnalysis` contract moved too**, taking the count from 9 to **1**. `RuleAnalysis` is an algorithm over interfaces with one adapter per layer (`ModelRuleFactory` over biological rules, `MathRuleFactory` over `ParticleJumpProcess`es) and references no concrete biological type — it was simply *declared* in `org.vcell.model.rbm`. It and `RuleAnalysisReport` now live in `cbit.vcell.math`, the layer both adapters can legally reach.

`getNFSimXML` (255 lines) deliberately did **not** move with it — it generates a solver input format, so keeping it beside the contract would reintroduce the P4 violation just fixed for the `.lngv` writers. It moved into `NFsimXMLWriter`, where its only live caller already was. Its other caller, `RulebasedTransformer.compareOutputs`, was deleted: unreachable (every call site commented out), and the only thing that would have forced a new biological→solver dependency.

**`RbmUtils` split too, taking the count to zero.** Its paired biological/math overloads meant roughly half of a 2191-line class in a biological package was really a math-side BNGL writer; the eight math-typed methods are now `cbit.vcell.math.ParticleBnglStringWriter`. The halves separated cleanly because they were only ever related by name — the biological overloads take a `Model`, a `Structure` or a `CompartmentMode`, none of which mean anything in the math layer.

Not a P4 violation: the `RuleAnalysis` contract is itself expressed in BNGL (`getReactionBNGLShort()`, `toBngl()`, `getMolecularTypeBNGL()`), so the math adapter is required to produce it, and BNGL is a modelling language rather than one solver's input format.

**`cbit.vcell.math` now imports nothing from the biological or solver layers — 18 violations at the start of this branch, zero now**, and `MathNamespaceSeparationTest`'s accepted list is empty.

## 3. Every math scalar is an `Expression` (P3)

`bondLength`, `radius` and `diffusionRate` were raw doubles; the Langevin math classes now have **none**. The rest of the layer was already uniform — `ParticleProperties`, `MacroscopicRateConstant`, `JumpProcess`, `PdeEquation`, `Constant` and `Function` carry no raw doubles at all.

They now compare via `ExpressionUtils.ExpressionEquivalencePredicate` (1e-9 relative / 1e-12 absolute) instead of exact `Double.compare`, which was **stricter than every neighbouring class** and would have reported last-bit drift as a difference.

VCML and XML write the infix **unflattened**: old documents hold float literals, new ones hold expressions that happen to be float literals, and both read back. `LangevinMathMapping` wraps the biological doubles at the boundary; `LangevinLngvWriter` evaluates with `flatten().evaluateConstant()` and throws `SolverException` naming the site and the offending infix rather than writing a wrong number.

## 4. Comparison

**Identical.** All three stubs implemented. `internalLinkSpec` was also uncompared and wasn't in the original survey — it reaches the solver alongside `is2D`, so it's included, compared by site-name pairs and **directed**, since springs are written to the solver in stored order.

The **asymmetry had to be fixed in the parents**: `ParticleMolecularType`, `ParticleMolecularComponent` and `ParticleJumpProcess` guarded with `instanceof`, which a subclass satisfies, so `plain.compareEqual(langevin)` was true while the reverse was false — and `Compare.isEqual(List,List)` only evaluates `v1[i].compareEqual(v2[i])`, making list comparison depend on argument order. Now exact-class, which is safe because the class is decided by the document tag and round-trips stably.

**Equivalence had a hole of the same kind, one tier up.** `compareEquivalentCanonicalMath` never compared `particleMolecularTypes` — canonicalisation flattens equations over state variables and neither compares nor preserves them, so a changed site radius was reported *equivalent*, keeping the parent reference and leaving stale results attached. Now compared, under `MathDifferent_DIFFERENT_PARTICLE_MOLECULAR_TYPES`.

### Differences

`MathCompareResults` now carries a list of `Difference` (`path`, `attribute`, `value1`, `value2`), attached by `testEquivalency` when maths are not equivalent:

```
molecularType 'A' / site 's0': radius '1.0' vs '99.0'
```

`compareEqual` remains the only definition of difference — the walker asks it whether two objects differ and only descends to *name* the attribute once the answer is already yes, so it cannot produce a false "no differences". If it can't identify the attribute it still reports the object as differing, with `attribute = "(not identified)"`. `toDatabaseStatus()` is untouched, since `MathVerifier` writes it to a 255-character column.

Colour sits exactly on the boundary — persisted but never sent to the solver — and is handled by a lenient `compareEqual(obj, bIgnoreDisplayAttributes)` overload following the existing `bIgnoreMissingDomain(s)` family, rather than a parallel field walker that could drift from `compareEqual`. So a colour change **is** a difference for *identical* (the edit saves) and **is not** for *equivalent* (results survive).

## 5. Two latent round-trip bugs, fixed first

Comparing a field that doesn't survive a save manufactures false differences, so the round trip was audited before enabling comparison:

- the VCML reader used `Boolean.getBoolean(token)` — which reads a **system property** of that name, not the token — so `is2D` always read back `false`; and `is2D` had no XML path at all, though its tag was declared;
- `bondLength` was parsed unconditionally with the failure swallowed (`// do nothing, the token was '-'`). That catch was load-bearing for the non-BINDING sentinel but also hid genuinely malformed values; it now guards on subtype, mirroring how `TransitionCondition` is read.

## 6. Regression corpus

`MathGenCompareTest` compares regenerated math against the math **stored in each document**, and had **no SpringSaLaD coverage at all**. The public models carrying Langevin math are now in it.

Their stored math wasn't usable as a baseline as-is: SpringSaLaD was under active development and fixes landed after several were saved, so 5 of 33 cases reported `DIFFERENT_PARTICLE_PROPERTIES` — four of them written by the same build (`Alpha_Version_7.5.0_build_101`). Rather than pin each model to whichever version saved it, math was **regenerated for all of them with a recent pre-PR VCell** (`a9d3049b8d`) and the documents re-saved; all 5 differences vanish, confirming stale storage rather than defects. Documents are stamped `Regenerated_baseline_a9d3049b8d`.

Deduplicated by top-level `BranchId` — a *copy* starts a new lineage, successive saves share one — giving **14 documents / 14 lineages / 25 cases**. Four applications still fail validation and are registered in `knownMathGenerationFailures` with the specific reason, keeping the rejection covered.

## 7. Evidence this didn't change generated math

`LangevinMathMapping` is where a mistranslation would land, and comparison tests can't catch one — both sides would be equally wrong. So math was generated for every SpringSaLaD model on the pre-separation commit and diffed: **byte-identical**. The baseline is therefore pre-PR output that this branch reproduces exactly, not the PR marking its own homework.

Each guard was also verified to fail when it should: injecting a `TRANSITION`→`ALLOSTERIC` mistranslation fails the corpus; removing the `is2D` XML write fails the serialization test; adding a biological or solver import fails the boundary test.

## Testing

```
mvn -o test -pl vcell-core -Dgroups=Fast                                  # green
mvn -o test -pl vcell-core -Dgroups=MathGen_IT -Dtest=MathGenCompareTest  # 705 tests, 0 failures
```

⚠️ `MathGenCompareTest` is `@Tag("MathGen_IT")` and so runs only in `regression.yml`, **which an admin-merge bypasses along with the merge queue**. These are run locally on each push; the per-field comparison tests are `Fast`-tagged and do gate every push.

## Outstanding

- resurrecting the BioNetGen cross-check as a maintained tool — deleting `compareOutputs` removed the only validation of our NFSim XML against BioNetGen's. It was unreachable, so nothing got worse, but it is worth doing deliberately; noted in `docs/nfsim-abstractions.md`
- baselining the `.lngv` solver input as the equivalence-tier oracle
- four models that fail validation: whether the models or the validation should change is a modelling decision, not addressed here


jcschaff:
Pushed five further commits, following the design review. The layer rules that drove them are now written down in `docs/architecture-layers.md`.

**P4 — solver formats out of the math namespace.** `LangevinParticleMolecularComponent.writeType` / `.writeSite` emitted `.lngv` syntax from inside `cbit.vcell.math`, and were the only reason it imported `org.vcell.solver.*`. Both moved to `LangevinLngvWriter`. The guard test enforced only half the rule — it forbade biological imports but not solver ones, so it would not have caught this; `FORBIDDEN_PACKAGES` now covers both directions, verified against an injected import.

**P3 — every math scalar is an `Expression`.** `bondLength`, `radius` and `diffusionRate` were raw doubles; the Langevin math classes now have none. They compare via `ExpressionUtils.ExpressionEquivalencePredicate` (1e-9 relative / 1e-12 absolute) instead of exact `Double.compare`, which was stricter than every neighbouring class — `ParticleProperties`, `MacroscopicRateConstant`, `JumpProcess` and `PdeEquation` carry no raw doubles at all. VCML and XML write the infix unflattened, so old documents holding literals and new ones holding expressions both read back. `LangevinMathMapping` wraps the biological doubles at the boundary; the solver writer evaluates with `flatten().evaluateConstant()` and throws `SolverException` naming the site and offending infix rather than writing a wrong number.

**The equivalence tier compared no molecular types at all.** Canonicalisation flattens equations over state variables, so it neither compares nor preserves the molecules the particle solvers move — a changed site radius was reported as *equivalent*, meaning `SimulationVersion.parentSimulationReference` was kept and results computed with the old radius stayed attached to the edited math. Now compared, under `MathDifferent_DIFFERENT_PARTICLE_MOLECULAR_TYPES`.

Colour sits on the boundary and is handled by a lenient `compareEqual(obj, bIgnoreDisplayAttributes)` overload, following the existing `bIgnoreMissingDomain(s)` family rather than a parallel field walker that could drift from `compareEqual`. So a colour change is a difference for *identical* (the edit saves) but not for *equivalent* (results survive).

**Two latent bugs found on the way**, both of which would have surfaced as false differences:
- the VCML reader used `Boolean.getBoolean(token)` — which reads a *system property* of that name, not the token — so `is2D` always read back `false`;
- `bondLength` was parsed unconditionally with the failure swallowed (`// do nothing, the token was '-'`), which was load-bearing for the non-BINDING sentinel but also hid genuinely malformed values. It now guards on subtype, mirroring how `TransitionCondition` is read.

**Testing.** vcell-core Fast group green; `MathGenCompareTest` **705 tests, 0 failures** across the whole corpus, run locally since `MathGen_IT` only runs in `regression.yml`, which an admin-merge bypasses.

One test worth calling out: `anEquivalentExpressionForTheSameValueIsNotAChange` asserts that `r` and `r + 0` stay equivalent. It would have failed under the previous exact `Double.compare`, and is the concrete payoff of holding math scalars as expressions.

Still outstanding: the `MathCompareResults` differences list is structure-only and not yet populated — `getDifferences()` returns empty, so an empty list means "nothing recorded", not "no differences".

Issue #1879, opened 2026-08-10
Field viewer: VTK-based visualization for moving boundary problems (mbsolver)


jcschaff:
Direction note from the #1859 initiative (browser field viewer, shipped through 8.0.9.01/8.0.10.01): extend the viewer to **MovingBoundary solver** results, as a step toward the browser viewer covering all spatial solvers (and eventually de-emphasizing the Java-based PDE viewer).

## Why this is closer than it looks

The protocol and code were built anticipating moving geometry:

- `/grid` is deliberately **not** documented as immutable — it is "the geometry for this dataset *at this time*", served with an ETag + short `max-age` instead of `immutable`, exactly so a per-timestep mesh can be re-requested cheaply (`FieldViewerServer.handleGrid` javadoc).
- `/field` carries the `geometryId` its values belong to, so values can never be silently paired with the wrong frame's mesh; the viewer already checks it (today it throws on mismatch — for moving boundaries it should trigger a `/grid` refetch instead).
- The server-side mapping **already exists**: `org.vcell.vis.mapping.movingboundary` builds a per-time `VisMesh` from `MovingBoundarySimFiles` (HDF5), from the old VisIt pipeline; `VisPolygon` carries the solver index like the Cartesian paths (the 2D quad path just landed in #1868 via the `Cells` accessor).

## Sketch

- **Server**: a MovingBoundary-aware `DataSource` (registration from `PDEDataViewer` when the run is mbsolver) whose `grid(source, domain, time)` maps the requested timestep through the existing movingboundary mapping; `geometryId` gains the time (or ALE segment) qualifier the javadoc already reserves for this.
- **Viewer**: on time scrub, when `/field` returns an unknown `geometryId`, refetch `/grid` and rebuild (machinery exists — `rebuildGeometry`). The **deform pipeline does not apply**: the mesh is body-fitted already, so the smoothing slider is disabled for these runs (a third mode next to 3D-Cartesian and 2D-Cartesian). Color bar, axes, time plots and stats endpoints carry over; picking needs a general point-in-cell locate for non-Cartesian meshes (`vtkCellPicker` is marshalled in the bundle as the fallback).
- **Future**: per the design doc (§8B), FEniCSx-era ALE meshes amortize connectivity per ALE segment; the mbsolver's 2D per-frame geometry is small enough to ship whole frames first and optimize later.

Related: #1859 (feature train), #1867 (color-scale modes), `docs/3d-renderer-design.md`.

[...]

PR #1890, opened 2026-08-11
Field viewer: Chombo solutions via the VTU seam (static embedded-boundary meshes, mixed cells)

jcschaff:
Adds Chombo solutions (both 2D and 3D chombo solvers — retired, but their stored results remain viewable) to the field viewer through the same VTU seam as MovingBoundary (#1879 / PR #1883), with one new distinction: `VtuMode.STATIC` (one embedded-boundary mesh, values vary per time) vs `TIME_VARYING` (a mesh per saved time). The scrub is geometryId-driven: the viewer refetches geometry only when the field's geometryId differs from the one on screen, so a static mesh ships once.

**The evidence below is REAL Chombo geometry** — not a mock. It comes from a fresh local run of BioModel 276459600 ("Solver_Suite_7_5", app "3D chombo", sim "chombo 3d": an 8×8×8 embedded-boundary sphere), executed with a from-source `VCellChombo3D_x64` build, read back through the actual `getEmptyVtuMeshFiles`/`getVtuMeshData` seam (Python VTK service included), and replayed byte-for-byte through this PR's viewer.

| real EB surface (RanC_nuc) | hemisphere after z-cut | face-on cross-section (fraction-0) |
|---|---|---|
| surface | hemisphere | cut face |

The cross-section is the embedded-boundary structure itself: whole voxels (fraction-0 = 1) inside, cut tetrahedra at the rim with their true partial volumes, a flat sharp cut, and no incorrect tet slices — the cut face tessellates the disk with no holes or overlaps. Consistency check: the mesh's integrated volume is 255.5 vs the analytic sphere's ≈248 (8³ resolution), and the mid-plane clip halves it exactly (127.7).

**Read the stray triangles in the first screenshot as a real mesh defect, not a rendering artifact — see #1895.** The cut face above tessellates cleanly, but the *volume mesh's boundary surface* does not: `ChomboMeshMapping.cropVoxels` clips each voxel independently against its own membrane facet plane, so neighbouring facets are never required to meet. The EB surface has 168 open edges (144 hairline seams 0.025 wide, plus a ~0.9 × 0.9 hole at each of the six poles), and 464 of the 1422 extracted surface faces are interior grid-plane walls exposed by tessellation mismatch — a full voxel's quad against a cut neighbour's two triangles. Through a pole hole you see one of those walls, coloured by its own cell's value; that is the bright green/cyan triangle. Verified as *not* the cause: coordinates and scale (raw doubled-index −1…15 → 0…10 exactly), volume consistency (Σ cell volume 255.45 vs Σ fraction·dx³ 256.41), and the mixed-cell reconstruction itself (zero inverted, degenerate or sliver tets, zero T-junctions, zero VTK_VOXEL corner-order violations). Volume-integrated statistics are unaffected. The defect predates this PR — it is the Chombo → VisMesh path, which the viewer renders faithfully.

### What landed since the hold

- **Physical coordinates**: a Chombo .vtu carries `ChomboMeshMapping`'s doubled-index coordinates (a VisIt-era convention). `chomboToPhysical` inverts that map as the mesh crosses the seam, so axes, lab-frame points and cell measures are microns.
- **Cut plane on body-fitted meshes**: the crop clips the solver mesh directly (no deform involved), exposing the solver's own interior cells; crop stats are labeled "body-fitted solver mesh".
- **Merged master**, generalizing #1891's lab-frame `/timeseries` and moving-domain `/stats` to VTU mode: STATIC locates the lab point once and reuses cell measures across times; the lab point gains an optional `z` for 3D.
- Mixed-cell VTUs (VTK_VOXEL + VTK_TETRA) render via per-cell `insertNextCell` (probed as marshalled in vtk.wasm ≥ v1.2.0).

### Dependencies / found along the way

- **Requires #1893** for the live server path: `VtkServicePython` has passed the output *directory* where the Python VTK service expects the .vtu *path* since July 2023, so server-side Chombo VTU generation was silently broken. (MovingBoundary was unaffected — its VTU writer is pure Java.)
- The chombo solver itself is not currently shipped anywhere (binaries absent from solver bundles and vcell-batch; see `docs/chombo-solver-notes.md` on this branch). This PR only needs the stored-data reading path.
- Legacy post-processing files (statistic name "mean") still abort unrelated data access in `Hdf5PostProcessor` — including for output written by today's solver; tracked separately.

### DIFF
--- a/docs/chombo-solver-notes.md
+++ b/docs/chombo-solver-notes.md
@@ -1,0 +1,141 @@
+# Chombo solver — status, run specification, and file formats
+
+Working notes gathered while adding Chombo support to the VTK-based field viewer
+(PR #1890). Chombo is a retired solver: stored results exist, but as of August 2026
+no deployed image or solver bundle can *run* it (details below). This document
+records what is needed to read its output today and to resurrect the solver later.
+
+## 1. Current availability (as of 2026-08)
+
+- **Source still exists**: `virtualcell/vcell-solvers` repo, directories
+  `VCellChombo/` (the VCell driver) and `Chombo/` (the LBNL library). The CMake
+  options `OPTION_TARGET_CHOMBO2D_SOLVER` and `OPTION_TARGET_CHOMBO3D_SOLVER`
+  default to **off**, and current CI does not enable them.
+- **No binaries anywhere current**:
+  - `localsolvers/` bundles (release `v0.0.44-dev4`: linux64/mac64/win64) contain
+    FiniteVolume, MovingBoundary, Sundials, VCellStoch, NFsim, smoldyn, langevin —
+    no `VCellChombo2D`/`VCellChombo3D`.
+  - `ghcr.io/virtualcell/vcell-batch:8.0.11.01` (the image the HPC Singularity/Apptainer
+    image is built from): `find / -iname '*hombo*'` is empty.
+  - `ghcr.io/virtualcell/vcell-solvers:latest` and `:v0.0.42-dev3`: none.
+- **Consequence**: a Chombo simulation submitted today would fail at executable
+  resolution (`SolverUtilities.getExes` → `ResourceUtil.findSolverExecutable`)
+  both client-side and on HPC. Resurrection means either building from source with
+  the two CMake options enabled (linux/amd64; Fortran + HDF5 toolchain) or
+  recovering binaries from a VCell 7.x-era image.
+
+## 2. How a Chombo run is specified
+
+- `SolverDescription.Chombo` — display name **"Chombo Standalone"** ("EBChombo,
+  Semi-Implicit (Fixed Time Step), Experimental"), KISAO:0000285. Features:
+  spatial, deterministic, region-size functions, Dirichlet-at-membrane-boundary,
+  parallel, fast systems.
+- `SolverExecutable.VCellChombo` maps to **two** executables: `VCellChombo2D` and
+  `VCellChombo3D`; `FVSolverStandalone` (the same Java driver used for the
+  FiniteVolume solver, via its `isChombo` flag) picks `getExes()[0]` for 2D and
+  `[1]` for 3D. `FiniteVolumeFileWriter` writes the input file with
+  `bChomboSolver` branching.
+- VCML: `<SolverTaskDescription Solver="Chombo Standalone">` containing a
+  `<ChomboSolverSpec>`:
+
+  ```xml
+  <ChomboSolverSpec>
+    <MaxBoxSize>32</MaxBoxSize>
+    <FillRatio>0.9</FillRatio>
+    <SaveVCellOutput>true</SaveVCellOutput>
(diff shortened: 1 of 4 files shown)


jcschaff:
### The slivers in the screenshots are real — and they are in the mesh, not in this PR

@jcschaff flagged that the screenshots above look wrong: stray triangles and thin slivers on the surface. They are real defects. They come from VCell's Chombo → VisMesh conversion, not from the viewer/server code here — the viewer draws faithfully what the `.vtu` contains. Filed as #1895.

Analysis of the actual `.vtu` behind these screenshots (907 points, 1778 cells: 1722 `VTK_TETRA` + 56 `VTK_VOXEL`), building the face map and keeping faces owned by exactly one cell — i.e. exactly what gets rendered as surface:

**The EB surface is not watertight: 168 open edges.** 144 are hairline seams **0.025 wide**, where adjacent cut cells' facets terminate at different points on their shared cell face (one at `x = 1.225`, its neighbour at `x = 1.250`). The other 24 form a **~0.9 × 0.9 hole at each of the six poles** of the sphere — which is precisely where the stray triangles appear in the surface screenshot. `ChomboMeshMapping.cropVoxels` clips each voxel independently against its own facet plane, so neighbouring facets are never required to meet; near a pole the plane is nearly parallel to the cell face, the cap cells collapse to one tet apiece, and the pole disk is left uncovered.

**Interior walls are exposed.** 464 of the 1422 rendered faces lie on *interior* grid planes — 341 of 560 area units, against only 201 for the sphere's own surface. A full voxel emits one quad on a shared face, its cut neighbour emits two triangles, nothing pairs, both get extracted. 153 area units are present from both sides (coincident double walls, z-fighting). Those walls, coloured by their own cells' values and seen through the pole holes, are the bright green/cyan triangles.

**What is not wrong**, since it bears on this PR's own code:

- **Coordinates and scale are correct.** Raw doubled-index coords span exactly −1…15 → 0…10 physical, so `chomboToPhysical` is right. Σ|cell volume| = 255.45 against Σ fraction·dx³ = 256.41 (0.4 %); Chombo's own EB volume is 4.4 % below the analytic 268.08 at 8³, as expected.
- Zero inverted tets, zero degenerate/sliver tets (V/L³ ≥ 1e-3 for all 1722), zero T-junctions, zero `VTK_VOXEL` corner-order violations, no face with more than two owners — so the mixed-cell `insertNextCell` path is reconstructing the grid correctly.
- The fat blue crescents at the rim of the cut face are correct: a cut cell's cross-section is widest exactly at the equator.

**Correction to the PR body above:** the claim that the cross-section shows "no holes or overlaps" holds for the cut face itself, but not for the volume mesh's boundary surface — it has both (cracks at the poles, doubled interior walls). I'll amend the body if we keep these screenshots.

Volume-integrated statistics are unaffected — the cell decomposition is sound; only the inter-cell agreement of boundary facets is missing.

PR #1901, opened 2026-08-12
Make the BMDB nightly pass or fail against its documented baseline

jcschaff:
The nightly posted a thousand-line report to Slack and was green regardless of what it found. Nobody reads it, and nothing enforces it.

## Why it never failed

Three causes, none of them a missing feature:

- `TestReportCommand.call()` ended in an unconditional `return 0`
- the only alarm could never fire — `grep -q PATTERN "$(cat report.md)"` passes the file's **contents** as the filename. It already carried a `# Not currently working` comment
- nothing else in the job could fail (no `continue-on-error` involved — nothing ever exited non-zero)

## The comparison already existed

`vcell-cli/src/main/resources/test_cases.ndjson` is the documented baseline — 1072 `SYSBIO_BIOMD` cases, 893 PASS + 179 FAIL, exactly the report's "Historical" column — and `OmexTestReport` already classified every deviation. It simply had no way to say so. This gives it one.

## What now fails a run

| condition | meaning |
|---|---|
| **changed** | an undocumented failure, a documented failure that has started passing, or a documented failure that now fails a *different* way |
| **stopped running** | a documented case that did not execute |
| **undocumented** | an execution with no baseline entry |

A newly *passing* case fails the run too: the baseline is now wrong, and left alone it masks the next real regression.

**"Stopped running" is the one that matters most**, because it reads as good news. Last night is the example: passes 893→889 and failures 179→172 looked like an improvement, but the entire delta was 11 models that did not run (4 passes, 6 `SOLVER_FAILURE`, 1 `TOO_SLOW`). Nothing actually improved.

## Those 11 turned out to be deliberate

They are exactly the workflow's own hard-coded skip list at line 179 — `235|255|595|711|1008|1013|1022|1025|1030|1035|1042`, commented *"too slow or otherwise misbehaved"* — never reconciled with the baseline. The gate would have gone red every night for an intentional exclusion.

`SKIP` already means "known not to run", so documented-SKIP cases are exempt from "stopped running", and those 11 are marked `SKIP` with the reason recorded (11 lines changed, formatting preserved).

## Keeping the gate survivable

`--update-test-cases` writes the baseline the run implies — across **all** collections, so it stays diffable against the committed file rather than dropping the 110 rows this run doesn't cover. Accepting a legitimate change becomes a reviewable diff instead of a hand-edit of 1182 lines. That is what decides whether a hard gate survives contact.

## Verified against real data

Against last night's actual `exec_summary.ndjson` (1061 lines, downloaded from run 31574242739):

- reports `FAIL: 3 changed` and names them — `BIOMD0000000459/460/461`, `SBML_IMPORT_FAILURE → UNCATEGORIZED_FAULT`
- `--update-test-cases` produces a file differing in **exactly those 3 lines** out of 1182
- re-running against that file exits **0**, `PASS: no changes`, and the report's Historical column now agrees with Current (889/172 both sides) instead of the misleading 893/179

So the first run goes red for a genuine reason, with a 3-line diff to accept.

## Slack is now two or three lines

```
🔴 *BMDB nightly: FAIL* - 3 changed
Accept these results by committing the `test_cases.updated.ndjson` artifact over `vcell-cli/src/main/resources/test_cases.ndjson`.
<run details, full report and updated baseline>
```

Green is one line plus the link. The full report and the updated baseline are artifacts of the run.


jcschaff:
Investigated the 3 `SBML_IMPORT_FAILURE` → `UNCATEGORIZED_FAULT` changes this gate surfaces. They are **not a regression** — the failure moved forward: these models now import and simulate, and fail at results CSV export (`NonSpatialSBMLSimResults:95`) rather than at import (`SedMLImporter:595`). Timings agree: import failures median 2230ms (n=40), these three 2990ms, passes 3528ms (n=889).

Pushed `30aa3239cf`, which adds `SBML_RESULTS_MAPPING_FAILURE` (a plain `RuntimeException` from results export matched no rule, hence `UNCATEGORIZED_FAULT`) and points the three baseline entries at it. Underlying defect filed as #1905.

Note `failure_type` is computed by `summarize()` at execution time, so the new rule takes effect from the next nightly. Verified by re-typing those three executions in last night's real `exec_summary.ndjson` as `summarize()` will now record them — gate exits 0. Left as recorded it still exits 1, so the pass isn't vacuous.

PR #1924, opened 2026-08-13
Document servertype, the second service selector inside data

jcschaff:
Comment only — no behaviour change.

`data` selects a service twice, and the consolidated entrypoint made that easy to confuse by giving "which service" a new meaning:

| | |
|---|---|
| `$1 = data` | which of the five VCell services the container runs (new) |
| `$servertype` | which role SimDataServer plays: `ExportDataOnly` / `SimDataOnly` / `CombinedData` |

`servertype` is `SimDataServerMain`'s sole argument and is required — `args.length` must be 1 and the value goes through `SimDataServiceType.valueOf`, so anything else throws. All five overlays set it in `config/<overlay>/data.env`, and all five say **`CombinedData`**: one pod serving both simulation data and export data. The image default is deliberately the invalid `servertype-not-set`, so a deployment that forgets it fails loudly rather than silently picking a role.

It is also not the `export` deployment, which is the separate Quarkus `vcell-exporter` image.

Behaviour is unchanged: the entrypoint already passed `${servertype}` exactly as `Dockerfile-data-dev` did, and stage's `data` pod is running on the merged image right now with `servertype=CombinedData` from its ConfigMap.

### DIFF
--- a/.github/workflows/CI-full.yml
+++ b/.github/workflows/CI-full.yml
@@ -137,18 +137,6 @@ jobs:
             needs_maven: false
             needs_secrets: false
             build_args: "BUILD_COMMAND=build_prod"
-          - name: webapp-island
-            dockerfile: webapp-ng/Dockerfile-webapp
-            context: webapp-ng
-            needs_maven: false
-            needs_secrets: false
-            build_args: "BUILD_COMMAND=build_island"
-          - name: webapp-remote
-            dockerfile: webapp-ng/Dockerfile-webapp
-            context: webapp-ng
-            needs_maven: false
-            needs_secrets: false
-            build_args: "BUILD_COMMAND=build_remote"
           - name: db
             dockerfile: docker/build/Dockerfile-db-dev
             context: .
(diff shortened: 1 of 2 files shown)


jcschaff:
**Correction to this PR's description.** It said "Comment only — no behaviour change". That was wrong: the branch was cut from `ci/trim-images-and-retry-push` rather than from `master`, so merging it landed **three** commits, not one:

| commit | what it actually did |
|---|---|
| `3023882534` | Document servertype (the described change) |
| `12ef094135` | Push only the two new tags, not every tag — 54 → 30 registry writes per release |
| `742809c400` | Stop building vcell-webapp-island, vcell-webapp-remote and vcell-mongo (18 → 15 images), and retry pushes through GHCR's secondary rate limit |

The two CI commits were intended and discussed — they are the fix for the `tag-and-push` failure on the 8.0.15.01 release — but they should have been reviewed as their own PR, not merged under a docs-only banner.

**One ordering risk, now closed.** Dropping `vcell-mongo` from the build requires virtualcell/vcell-fluxcd#32, which stops `deploy.yaml` rewriting its tag; without it a deploy would request a mongo image that was never built. That companion landed 8 minutes later (10:45 → 10:53 EDT), and no release or deploy happened in the window — 8.0.15.01 was published at 13:39 UTC and deployed at 14:15 UTC, both before. So nothing was affected.

No revert needed; `ci/trim-images-and-retry-push` is redundant and is being deleted.

PR #1945, opened 2026-08-14
Stop re-downloading the same dependencies on every CI job

jcschaff:
Your hypothesis was right, and it is worse than "not threaded" — the downloads are **entirely repeated work**. Two independent causes, both measured on PR #1943's build job (maven `Total time 08:17`) whose dependency cache had been *restored successfully*.

### 1. The cache never updated

`setup-java`'s `cache: maven` saves only on a cache **MISS**:

```
Post setup java 17 with maven cache: Cache hit occurred on the primary key …, not saving cache
```

The first cache ever written was incomplete. Because the key changes only when a pom does, every run since restored that same incomplete snapshot and never saved an updated one. **1222 artifacts (811 poms, 409 jars) were downloaded on a cache hit** — and would be on every job of every run indefinitely.

Replaced with an explicit cache whose key always differs (so it always saves) plus `restore-keys` falling back to the newest previous one, so it converges on complete instead of freezing incomplete. **Only one job per workflow saves** — `build` in ci.yml, `maven-build` in CI-full.yml — so a run adds one entry, not four (or sixteen across the `docker-build` matrix).

### 2. Central was consulted last

Central is not declared in our pom at all, so it comes from the super-POM **after** every repository we declare. Nearly everything we depend on is in Central, so each artifact was looked up in five other repositories first:

| repo | attempts | served |
|---|---|---|
| central | 1207 | **1205** |
| terracotta | 1157 | 0 |
| scijava.public | 1157 | 0 |
| jitpack.io | 1157 | 0 |
| com.alfresco.repository.public | 1157 | 0 |
| lib-maven-repo | 1156 | 0 |

**7028 attempts produced 1222 artifacts** — roughly 5800 failed round trips per job. Declaring Central first makes the common case stop at the first repository; the others still serve the handful of artifacts genuinely only there, they are just not asked first. Verified in the effective POM: `central` is now first of six.

### Why they compound

Fixing the cache removes most downloads; fixing the order makes whatever remains cost one lookup instead of six.

### What I did not change

`ci.yml:11-13` documents that each job compiles independently on purpose ("reusing compiled output across jobs proved unreliable for inter-module test-runtime classpaths"). I left that alone. If you want to go further afterwards, `-T 1C` for parallel module builds is the contained next step — it speeds each job without sharing artifacts between them — but it deserves its own timed experiment since not every plugin is thread-safe.

### Measuring it

The first run on this branch still pays a cold-ish cache (the pom hash changes, so it falls back through `restore-keys`) and **writes a complete cache**. The run *after* it is the one to compare — expect the `Downloading from` count to collapse. Worth watching `maven Total time` on the build job over the next few merges rather than judging from one run, given the 3:24–8:17 spread I measured across today's runs on identical work.

### DIFF
--- a/.github/workflows/CI-full.yml
+++ b/.github/workflows/CI-full.yml
@@ -66,12 +66,25 @@ jobs:
           popd;\
         done
 
-    - name: setup java 17 with maven cache
+    - name: setup java 17
       uses: actions/setup-java@v4
       with:
         distribution: 'temurin'
         java-version: '17'
-        cache: 'maven'
+
+    # Not setup-java's `cache: maven`, which saves only on a cache MISS -- so the first cache
+    # ever written, incomplete, was restored unchanged by every later run and never updated.
+    # A key that always differs means it is always saved, and restore-keys falls back to the
+    # newest previous one, so the cache converges on complete. This job saves for the run;
+    # docker-build restores only.
+    - name: maven repository cache
+      uses: actions/cache@v4
+      with:
+        path: ~/.m2/repository
+        key: maven-${{ hashFiles('**/pom.xml') }}-${{ github.run_id }}
+        restore-keys: |
+          maven-${{ hashFiles('**/pom.xml') }}-
+          maven-
 
     - name: Maven build
       run: mvn --batch-mode clean install dependency:copy-dependencies -DskipTests=true
@@ -190,13 +203,23 @@ jobs:
 
     - uses: actions/checkout@v4
 
-    - name: setup java 17 with maven cache
+    - name: setup java 17
       if: ${{ matrix.image.pre_build }}
       uses: actions/setup-java@v4
       with:
         distribution: 'temurin'
         java-version: '17'
-        cache: 'maven'
+
+    # Restore only. This is a matrix job, so saving here would write one cache entry per image.
+    - name: maven repository cache (restore only)
+      if: ${{ matrix.image.pre_build }}
+      uses: actions/cache/restore@v4
+      with:
+        path: ~/.m2/repository
+        key: maven-${{ hashFiles('**/pom.xml') }}-${{ github.run_id }}
+        restore-keys: |
+          maven-${{ hashFiles('**/pom.xml') }}-
+          maven-
 
     - name: Download Maven artifacts
       if: ${{ matrix.image.needs_maven }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -162,12 +162,27 @@ jobs:
           poetry install
           poetry run python -m pytest
 
-      - name: setup java 17 with maven cache
+      - name: setup java 17
         uses: actions/setup-java@v4
         with:
           distribution: 'temurin'
           java-version: '17'
-          cache: 'maven'
+
+      # Not setup-java's `cache: maven`. That saves only on a cache MISS, so the first cache
+      # ever written -- which was incomplete -- was restored unchanged by every later run and
+      # never updated, because the key only changes when a pom does. One build measured 1200
+      # artifacts (811 poms, 409 jars) re-downloaded on a cache HIT, in every job of every run.
+      #
+      # A key that always differs means the cache is always saved, and restore-keys falls back
+      # to the newest previous one, so it converges on complete instead of freezing incomplete.
+      - name: maven repository cache
+        uses: actions/cache@v4
+        with:
+          path: ~/.m2/repository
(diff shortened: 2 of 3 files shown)


jcschaff:
Good catch — you were right, and I've reordered it: **`lib-maven-repo` first, then Central**, then the network repositories.

Three reasons, the second of which I only found by checking:

1. A miss in a `file://` repository is a filesystem stat, not a network round trip. Asking it first costs the ~1200 Central-hosted artifacts almost nothing, while the artifacts that are only vendored resolve with **no network call at all** — better than what I had, which made those three pay a Central 404 first.
2. **Vendoring should mean what it says, and there is a real overlap.** `edu.ucar:netcdf:4.3.22` is vendored *and* published on Central:

   ```
   vendored sha1: a9b0be3ddf020c3aebf6498fcfc3ef007805cec0   4591382 bytes
   central  sha1: a9b0be3ddf020c3aebf6498fcfc3ef007805cec0   4591382 bytes
   ```

   Byte-identical, so my Central-first commit changed nothing in practice — but only by coincidence. With Central first the vendored copy would quietly stop being the one used, and that should not depend on the bytes happening to match. (`jai_core`, `jai-codec` and `udunits` return 404 from Central, so those were always going to fall through.)
3. It restores the original intent rather than reordering it as a side effect of the 404 fix.

The 404 problem is unaffected: Central still precedes alfresco, scijava, jitpack and terracotta, which is where the ~5800 wasted round trips came from.

Effective POM verified: `lib-maven-repo`, `central`, `com.alfresco.repository.public`, `scijava.public`, `jitpack.io`, `terracotta`. `mvn -o validate` passes.

jcschaff:
Corrected the pom comment, which was misleading — including in a way that shaped my own analysis earlier in this PR.

It said *"Vendored artifacts not available on Maven Central"*. Per Jim, they were vendored **for reliability, not availability**: the servers carrying them were flaky enough to fail builds, so a copy lives in the tree and compiling never depends on a third party being up.

That distinction decides what may be dropped later. On the old wording, `edu.ucar:netcdf:4.3.22` looks like a stale entry to clean up now that Central has it byte-identical — when availability was never the point. (`jai_core`, `jai-codec` and `udunits` do 404 from Central.)

It also gives the ordering a second justification I had not made. The remaining remote repositories — alfresco, scijava, jitpack, terracotta — **are** the flaky ones, and they were absorbing ~5800 lookups per job for artifacts none of them carry. Putting Central ahead of them reduces **exposure to those servers**, not just latency: an artifact found on Central is never asked of them at all. A flaky server that hangs rather than 404s currently stalls every build that needs a dependency it does not have.

Final effective order: `lib-maven-repo` → `central` → `com.alfresco.repository.public` → `scijava.public` → `jitpack.io` → `terracotta`. XML valid, `mvn -o validate` passes.

Three commits now:
1. cache that actually updates + Central ahead of the remote repositories
2. vendored file:// repository back in front of Central
3. the rationale correction above

jcschaff:
Answered by measurement — and only **one** repository turned out to be droppable.

The warm-cache CI logs could not settle this: an artifact already in the local repository is never looked up, so a repository can look unused when it is not. I ran two cold-cache `mvn clean install -DskipTests` builds against an empty local repository, one with alfresco and one without. **Both BUILD SUCCESS**, every artifact resolved from the same place either way:

| repository | artifacts served | what |
|---|---|---|
| central | 2465 / 2471 | everything else |
| maven-restlet | 15 | restlet 2.4.3 |
| **jitpack.io** | **13** | ome bioformats, and **our own `virtualcell/vcell-jsbml` fork** |
| lib-maven-repo | 9 | the vendored jars |
| **scijava.public** | **6** | cisd jhdf5, cisd base, saalfeldlab n5 |
| ej-technologies | 2 | install4j-runtime |
| **terracotta** | **1** | ehcache `sizeof-agent` pom — **404 on Central** |
| **alfresco** | **0** | — |

So jitpack, scijava and terracotta all earn their place; the warm-cache figures in my earlier comment (0 served for each) were an artifact of the cache, exactly the trap I should have expected. Only **alfresco** is unused, and it has been there since the original single-maven-project reorganisation.

**Honest accounting of the win:** removing alfresco saves little on its own — 4285 lookup attempts became 4253 — because with Central ordered ahead of it, it was already rarely reached. The **ordering** is what removed the waste. What the removal buys is one fewer third-party server on the critical path, which matters for the same reason the jars were vendored: a flaky server that hangs rather than 404s stalls the build.

Commented rather than deleted, so it is one line to restore if something outside the default profile set needs it.

Active repositories are now: `lib-maven-repo`, `central`, `scijava.public`, `jitpack.io`, `terracotta`.

PR #1993, opened 2026-08-18
Don't collect rows a save is still in the middle of writing

### LINKED ISSUE #1961: Database cleanup aborts the whole run on ORA-02292 a few times a day (concurrent-save contention, ~2% of runs)
## What is happening

`DatabaseServer`'s cleanup thread fails a few times a day with:

```
ORA-02292: integrity constraint (VCELL.SYS_C008209) violated - child record found
  at DBBackupAndClean.executeUpdate(DBBackupAndClean.java:1067)
  at DBBackupAndClean.cleanRemoveUnreferencedSimulations(DBBackupAndClean.java:836)
  at DBBackupAndClean.cleanupDatabase(DBBackupAndClean.java:1576)
  Caused by: SQL = DELETE FROM vc_simulation WHERE vc_simulation.id IN ( SELECT id FROM vc_simulation MINUS ...
```

Observed on **prod** (2 in 16 h) and **dev** (3 in 24 h). Not a regression: `DBBackupAndClean` has not changed since 2023, and prod showed this within hours of moving to 8.0.23 from a release three months old.

## Why it is not a schema fault

The delete already excludes exactly the rows protected by the two foreign keys that lack `ON DELETE CASCADE`:

```sql
DELETE FROM vc_simulation WHERE id IN (
    SELECT id FROM vc_simulation
    MINUS SELECT simRef FROM vc_biomodelsim              -- FK, no cascade
    MINUS SELECT simRef FROM vc_mathmodelsim             -- FK, no cascade
    MINUS SELECT versionParentSimRef FROM vc_simulation WHERE ... )
```

The other three children — `ResultSetMetaData`, `ResultSetExports`, `SimStat` — are `ON DELETE CASCADE` and cannot be the blocker. The cleanup runs every 15 minutes, so ~64 runs in the 16 h window with 2 failures: **~3%**. A missing exclusion would fail every time.

So a failure means a `vc_biomodelsim` (or `vc_mathmodelsim`) row was committed **between the subquery's read-consistent snapshot and the constraint check**.

## Two candidate mechanisms — evidence for neither

**A. A concurrent model save.** A user saves a BioModel that links a simulation the cleanup had just judged unreferenced. `ServerDocumentManager` inserts `vc_biomodelsim` rows as part of a save, so the window is real.

**B. Sites contending with each other.** dev, stage and prod **all point at the same database** — `jdbc:oracle:thin:@vcell-oracle.cam.uchc.edu:1521/ORCLPDB1` — and each runs its own cleanup thread every 15 minutes. Effective cadence across the three is a cleanup starting every ~5 minutes, unsynchronised, each with its own snapshot. A save on *any* site can trip *any* other site's in-flight cleanup.

These are not exclusive and both may contribute. **How to tell them apart:** correlate failure timestamps across the three `db` services against save traffic. Clustering with other sites' cleanup activity points at B; clustering with saves points at A. At 2–3 events a day this is slow evidence to collect.

## Consequences

**Each failure aborts the rest of the run.** `cleanRemoveUnreferencedSimulations` is step 2 of 8 in `cleanupDatabase`; the exception propagates, so unreferenced math descriptions, geometries, simulation contexts, models and software-version rows are skipped that run. The thread catches and retries in 15 minutes, so at 3% the effect is small — but the failure is broader than the one delete.

**`vc_simdelfromdisk` accumulates rows that no longer describe reality.** `executeUpdate` commits per statement, so the INSERT recording simulations for disk deletion commits *before* the DELETE fails, and the rollback covers only the DELETE. Three sites inserting the same rows compounds it.

*Not currently harmful*: the consumer re-validates with `AND simid not in (select id from vc_simulation)`, and it is a manual entry point (`deleteSimsFromDiskMethod1/2`), not something a service runs. It is a trap for whoever writes the next consumer, though — that table's contents are not a valid work list on their own.

**The interleaving we see is the lucky one.** The foreign key caught it. In the opposite order — the cleanup's DELETE committing just before a save's link insert — nothing catches it and the save fails on a missing parent instead. No evidence that has happened; the window is the same.

## Proposed first step

Not a SQL change. **Give the cleanup a flag so exactly one site runs it** (`vcell.db.cleanup.enabled`, opt-in, prod only). That removes two-thirds of the snapshot windows and the lock contention without touching the delete logic, and it makes mechanism B testable by elimination: if failures continue with a single site running the cleanup, the cause is A.

Worth deciding separately, once that is in place:

- whether `cleanRemoveUnreferencedSimulations` should catch `ORA-02292` on that statement and continue to steps 3–8 rather than aborting the run, since losing the whole cleanup to one contended row is disproportionate;
- whether the insert into `vc_simdelfromdisk` and the delete should share a transaction, so the work list cannot outlive the delete that justified it.

### LINKED ISSUE #1992: Database cleanup deletes rows a save is still writing → ORA-02291 breaks the save (save-side twin of #1961)
## Summary

The 15-minute database cleanup sweep deletes rows that a save is still in the middle of writing, breaking the save with `ORA-02291: parent key not found`. This is the save-side twin of #1961, which is the same race seen from the sweep's side (`ORA-02292: child record found`).

It fired a Better Stack alert on **2026-08-14 20:09 EDT** (`VCell Release Health - Sim`, prod, self-cleared in 2 min).

## Why "unreferenced" does not mean "garbage"

`ServerDocumentManager.saveBioModel` is **not one transaction**. Each child is committed by its own `DBTopLevel.insertVersionable` call (`DBTopLevel.java:1325`, `con.commit()`), and the `vc_biomodelsimcontext` / `vc_biomodelsim` link rows are only written in a **later** transaction, in `BioModelDbDriver.insertBioModelMetaData`.

For the few hundred milliseconds in between, the child row is committed and referenced by nothing — indistinguishable, to `DBBackupAndClean`, from an orphan left behind by a delete:

```sql
DELETE FROM vc_simcontext
 WHERE id NOT IN (SELECT simContextRef FROM vc_biomodelsimcontext)
```

There is no age guard on any of the five unreferenced-row deletes.

## Evidence — two prod incidents, both with the sweep mid-run

**2026-08-14 (prod `api` 8.0.23.01)**

```
runsim monitor thread  ERROR  ORA-02291: integrity constraint (VCELL.SYS_C008211) violated
  SQL = INSERT INTO vc_biomodelsimcontext (id,biomodelRef,simContextRef)
        VALUES (322090139, 322090135, 322090117)
  BioModelDbDriver.insertSimContextEntryLinkSQL(BioModelDbDriver.java:510)
  ServerDocumentManager.saveBioModel(ServerDocumentManager.java:1465)
  HealthService.runsimLoop(HealthService.java:205)
```

The missing parent, `vc_simcontext` **322090117**, is a row that same save had committed ~150 ms earlier — its id lands immediately after that save's last `vc_diagram` insert (322090116).

| time (UTC) | pod | event |
|---|---|---|
| 00:06:12.373 | `dev/db-7867d5b59d-jwd5c` | `Database Cleanup Thread` — `DBBackupAndClean` statement |
| **00:06:12.509** | `prod/api-5fd48444c5-6p7s6` | **ORA-02291 on the link insert** |
| 00:06:12.619 | `dev/db-7867d5b59d-jwd5c` | next `DBBackupAndClean` statement |

**2026-08-07 (prod `api` 8.0.0.03 — predates the 8.0.23.01 deploy, so not a regression)**

The sweep won a step earlier: `INSERT INTO vc_simcontext` failed on `SYS_C008289` (`mathRef`/`geometryRef`) at 07:06:28.559 while dev's cleanup ran statements at 07:06:28.310 and .773 — i.e. `DELETE FROM vc_math …` and `DELETE FROM vc_geometry WHERE dimension = 0 …`.

Those are the only two `ORA-02291` occurrences in prod in 14 days of Loki retention, and a cleanup sweep was in flight for both.

## Why the health check is the usual victim

Prod `api` runs 3 replicas, each with its own `HealthService` runsim loop on a 5-minute cycle — ~36 save → run → `deleteBioModel` cycles per hour. Each cycle orphans a simcontext/math/geometry/model, so the health check is both the largest producer of the garbage this sweep exists to collect and its most frequent casualty. Real user saves have the same window and, being slower, a wider one.

## Relationship to the one-site fix

`1a27cd646a` ("Let one site own the database cleanup", 8.0.24.01) added `vcell.db.cleanup.enabled` and now has dev and stage DISABLED with prod owning the sweep — verified in the running pods. That removes the cross-site amplification that caused both incidents above, but **not the race**: prod's own sweep still deletes rows prod's own saves are mid-way through referencing, roughly a third as often.

## Fix

Require a row to have been unreferenced for `CLEANUP_MIN_AGE_HOURS` (1) before collecting it, applied to all five deletes and to the report queries that must stay consistent with them — including the `vc_simdelfromdisk` insert-select that feeds disk cleanup:

```sql
DELETE FROM vc_simcontext
 WHERE vc_simcontext.id NOT IN (SELECT vc_biomodelsimcontext.simContextRef FROM vc_biomodelsimcontext)
   AND vc_simcontext.versionDate < SYSDATE - INTERVAL '1' HOUR
```

The guard belongs in `cleanRemoveUnreferencedSimulations`, **not** in the shared `getSelectUnreferencedSimKeySQL`, which `SimulationDispatcher:636` also uses to abort active jobs.

Branch `fix/cleanup-age-guard`, commit e96a9d6e90.

Verified:
- `mvn compile -pl vcell-server -am` clean.
- Captured every statement the patched `cleanupDatabase` emits, in both dialects, via a stubbed `Connection` — the guard is present on all five deletes and their matching selects, with `SYSDATE` for Oracle and `CURRENT_TIMESTAMP` for Postgres.
- Ran all 15 generated Postgres statements against a real Postgres against a stub schema: all accepted.
- Behavioural check on that Postgres: an unreferenced row 2 h old is collected, an unreferenced row written "just now" survives. Negative control — the pre-patch statement deletes both.

Not verified: execution against Oracle (`SYSDATE - INTERVAL '1' HOUR` is checked by construction only).

## Out of scope, noted while here

- `cleanRemoveUnReferencedSotwareVersions` is left unguarded: `vc_softwareversion` rows are inserted in the same transaction as their versionable (`insertVersionableInit`), so there is no window, and the table has no `versionDate`.
- `SimulationDispatcher`'s "unreferenced simulation" abort path shares the same definition of unreferenced. The window there is much narrower (a job can only be active after its save returned), but it is the same idea.

jcschaff:
Fixes #1992. Should also close #1961, which is the same race seen from the sweep's side.

## The race

`DBBackupAndClean.cleanupDatabase` — run every 15 minutes by `DatabaseServer.DatabaseCleanupThread` — collects rows that no document references. But "unreferenced" does not mean "garbage", because `ServerDocumentManager.saveBioModel` is **not one transaction**: each child (geometry, math description, model, simulation context, simulation) is committed by its own `DBTopLevel.insertVersionable` call, and the `vc_biomodelsimcontext` / `vc_biomodelsim` link rows are only written in a **later** transaction.

For the few hundred milliseconds in between, the child is committed and referenced by nothing — indistinguishable, to these queries, from an orphan left behind by a delete. Collecting it there breaks the save in progress with `ORA-02291: parent key not found`; when the race goes the other way the sweep aborts instead with `ORA-02292: child record found` (#1961).

Prod lost that race twice in the last 14 days of Loki retention, and a cleanup sweep was in flight for both:

```
2026-08-14 20:06:12  INSERT INTO vc_biomodelsimcontext (id,biomodelRef,simContextRef)
                     VALUES (322090139,322090135,322090117)
                     ORA-02291 (VCELL.SYS_C008211)   <- vc_simcontext 322090117 was
                                                        committed by this same save ~150 ms earlier
2026-08-07 03:06:28  INSERT INTO vc_simcontext ...
                     ORA-02291 (VCELL.SYS_C008289)   <- mathRef/geometryRef, swept a step earlier
```

The 2026-08-14 one raised the Better Stack `VCell Release Health - Sim` alert. Both predate or are unrelated to any recent change here — `1a27cd646a` (8.0.24.01) removed the cross-site amplification by giving prod sole ownership of the sweep, but not the race itself.

## The change

Require a row to have been unreferenced for `CLEANUP_MIN_AGE_HOURS` (1) before collecting it:

```sql
DELETE FROM vc_simcontext
 WHERE vc_simcontext.id NOT IN (SELECT vc_biomodelsimcontext.simContextRef FROM vc_biomodelsimcontext)
   AND vc_simcontext.versionDate < SYSDATE - INTERVAL '1' HOUR
```

Applied to all five unreferenced-row deletes and to the report queries that must stay consistent with them — including the `vc_simdelfromdisk` insert-select, which would otherwise queue simulations for disk deletion that the `DELETE` no longer removes.

Two deliberate boundaries:

- The simulations guard goes in `cleanRemoveUnreferencedSimulations`, **not** in the shared `getSelectUnreferencedSimKeySQL`. `SimulationDispatcher:636` uses that query to abort active jobs; its behaviour is unchanged.
- `cleanRemoveUnReferencedSotwareVersions` is left alone. `vc_softwareversion` rows are written in the same transaction as their versionable (`insertVersionableInit`), so there is no window, and the table has no `versionDate`.

Garbage an hour old is still garbage; a row a few hundred milliseconds old is a save in flight.

## Verification

- `mvn compile -pl vcell-server -am` clean.
- Drove the patched `cleanupDatabase` through a `java.lang.reflect.Proxy` stub `Connection` to capture every statement it emits in both dialects. The guard is present on all five deletes and their matching selects, with `SYSDATE` for Oracle and `CURRENT_TIMESTAMP` for Postgres.
- Ran all 15 generated Postgres statements against a real `postgres:15-alpine` on a stub schema — all accepted.
- Behavioural check on that Postgres: an unreferenced row 2 h old is collected; an unreferenced row written "just now" survives. Negative control — the pre-patch statement deletes both.

**Not verified:** execution against Oracle. `SYSDATE - INTERVAL '1' HOUR` is checked by construction only.

https://claude.ai/code/session_01HYdenYzMw35USHDQEATGVq

### DIFF
--- a/.github/workflows/regression.yml
+++ b/.github/workflows/regression.yml
@@ -1,6 +1,7 @@
 name: Regression
 
-# Heavy integration suites (MathGen/SBML/SEDML_*/BSTS). Deliberately NOT run on
+# Heavy integration suites (MathGen/SBML/SEDML_*/BSTS) plus the Oracle-dialect
+# check (Oracle_IT). Deliberately NOT run on
 # ordinary pushes - that is what keeps the fast lane (ci.yml) fast.
 #
 # Runs when:
@@ -51,6 +52,7 @@ on:
           - SEDML_SBML_IT
           - SEDML_VCML_IT
           - BSTS_IT
+          - Oracle_IT
       include-slow:
         description: 'Also run the slow cases excluded from the merge gate'
         type: boolean
@@ -98,6 +100,12 @@ jobs:
               "SEDML_VCML_IT": {"count": 2, "module": CORE},
               "SBML_IT":       {"count": 1, "module": CORE},
               "BSTS_IT":       {"count": 1, "module": "vcell-cli"},
+              # Not a model-suite sweep: one Oracle container proving the database
+              # cleanup sweep's SQL is accepted by the database it actually runs
+              # against in production. Its PostgreSQL twin is in the fast lane
+              # (vcell-rest), but Oracle needs its own ~2 GB image, so it is gated
+              # here rather than pulled on every push. See issue #1992.
+              "Oracle_IT":     {"count": 1, "module": "vcell-server"},
           }
           include = []
           for group, cfg in groups.items():
--- a/vcell-rest/src/test/java/org/vcell/restq/DatabaseCleanupSqlTest.java
+++ b/vcell-rest/src/test/java/org/vcell/restq/DatabaseCleanupSqlTest.java
@@ -1,0 +1,155 @@
+package org.vcell.restq;
+
+import cbit.vcell.modeldb.DBBackupAndClean;
+import cbit.vcell.resource.PropertyLoader;
+import io.quarkus.test.junit.QuarkusTest;
+import jakarta.inject.Inject;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+import org.vcell.restq.config.CDIVCellConfigProvider;
+import org.vcell.restq.db.AgroalConnectionFactory;
+
+import java.sql.Connection;
+import java.sql.PreparedStatement;
+import java.sql.ResultSet;
+import java.sql.Statement;
+import java.sql.Timestamp;
+import java.time.Duration;
+import java.time.Instant;
+
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+/**
+ * Exercises DBBackupAndClean.cleanupDatabase() -- the sweep the db service runs every 15
+ * minutes -- against a real database with the real VCell schema.
+ *
+ * This is the PostgreSQL half of the dialect coverage; the ORACLE branch of the generated
+ * SQL is covered by DatabaseCleanupOracleSyntaxTest in vcell-server.
+ */
+@QuarkusTest
+public class DatabaseCleanupSqlTest {
+
+	@Inject
+	AgroalConnectionFactory connectionFactory;
+
+	// well away from the key sequence any other test uses
+	private static final long USER_KEY   = 990000001L;
+	private static final long EXTENT_KEY = 990000002L;
+	private static final long GEOM_KEY   = 990000003L;
+	private static final long MODEL_KEY  = 990000004L;
+	private static final long OLD_SC_KEY = 990000005L;
+	private static final long NEW_SC_KEY = 990000006L;
+
+	@BeforeAll
+	public static void setupConfig(){
+		PropertyLoader.setConfigProvider(new CDIVCellConfigProvider());
+	}
+
+	@AfterEach
+	public void removeFixture() throws Exception {
+		Object lock = new Object();
+		Connection con = connectionFactory.getConnection(lock);
+		try (Statement stmt = con.createStatement()) {
+			stmt.executeUpdate("DELETE FROM vc_simcontext WHERE id IN ("+OLD_SC_KEY+","+NEW_SC_KEY+")");
+			stmt.executeUpdate("DELETE FROM vc_model WHERE id = "+MODEL_KEY);
+			stmt.executeUpdate("DELETE FROM vc_geometry WHERE id = "+GEOM_KEY);
+			stmt.executeUpdate("DELETE FROM vc_geomextent WHERE id = "+EXTENT_KEY);
+			stmt.executeUpdate("DELETE FROM vc_userinfo WHERE id = "+USER_KEY);
+			con.commit();
+		} finally {
+			connectionFactory.release(con, lock);
+		}
+	}
+
+	/**
+	 * Every statement the sweep issues has to be accepted by the database it is pointed at.
+	 * A dialect slip in DBBackupAndClean is invisible until the db service runs in production,
+	 * where it aborts the whole sweep.
+	 */
+	@Test
+	public void cleanupSweepIsAcceptedByTheDatabase() throws Exception {
+		Object lock = new Object();
+		Connection con = connectionFactory.getConnection(lock);
+		try {
+			StringBuffer log = new StringBuffer();
+			DBBackupAndClean.cleanupDatabase(con, log, connectionFactory.getDatabaseSyntax());
+			// cleanupDatabase() records each statement it ran; five deletes plus the
+			// version-parent updates and report queries.
+			assertTrue(log.indexOf("DELETE FROM vc_simcontext") >= 0, "sweep did not reach the simcontext delete");
+		} finally {
+			connectionFactory.release(con, lock);
+		}
+	}
+
+	/**
+	 * saveBioModel commits each child in its own transaction and writes the
+	 * vc_biomodelsimcontext link row in a later one, so a simulation context is briefly
+	 * committed and referenced by nothing. Collecting it there breaks the save with
+	 * ORA-02291 (issues #1992, #1961) -- the age guard is what stops that.
+	 */
+	@Test
+	public void ageGuardSparesRowsThatASaveIsStillWriting() throws Exception {
+		Object lock = new Object();
+		Connection con = connectionFactory.getConnection(lock);
+		try {
+			Instant now = Instant.now();
+			insertFixture(con, Timestamp.from(now.minus(Duration.ofHours(2))), Timestamp.from(now));
+
+			// neither simulation context is linked to a BioModel: by the sweep's own
+			// definition both are unreferenced.
+			assertTrue(exists(con, OLD_SC_KEY));
+			assertTrue(exists(con, NEW_SC_KEY));
+
+			StringBuffer log = new StringBuffer();
+			DBBackupAndClean.cleanupDatabase(con, log, connectionFactory.getDatabaseSyntax());
+
+			assertFalse(exists(con, OLD_SC_KEY), "an orphan two hours old should still be collected");
+			assertTrue(exists(con, NEW_SC_KEY), "a row a save is still writing must not be collected");
+		} finally {
+			connectionFactory.release(con, lock);
+		}
+	}
+
+	private boolean exists(Connection con, long simContextKey) throws Exception {
+		try (PreparedStatement stmt = con.prepareStatement("SELECT id FROM vc_simcontext WHERE id = ?")) {
+			stmt.setLong(1, simContextKey);
+			try (ResultSet rset = stmt.executeQuery()) {
+				return rset.next();
+			}
+		}
+	}
+
+	private void insertFixture(Connection con, Timestamp oldDate, Timestamp newDate) throws Exception {
+		exec(con, "INSERT INTO vc_userinfo (id,USERID,PASSWORD,EMAIL,FIRSTNAME,NOTIFY,insertDate,DIGESTPW)"
+				+ " VALUES (?,?,?,?,?,?,?,?)",
+				USER_KEY, "cleanupSweepTestUser", "x", "x@example.org", "cleanup", "N", newDate, "x");
+		exec(con, "INSERT INTO vc_geomextent (id,extentX,extentY,extentZ) VALUES (?,?,?,?)",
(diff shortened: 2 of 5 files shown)


jcschaff:
Added dialect coverage — the sweep had none.

**`DatabaseCleanupSqlDialectTest`** (vcell-server, `Fast`, no database) drives `cleanupDatabase()` through a `Connection` that records statements instead of executing them, and asserts the invariant that **no unreferenced-row delete goes out without the age guard**, in either dialect — along with the report query and the `vc_simdelfromdisk` hand-off that have to match it. `vc_softwareversion` is the documented exception (its rows are written in the same transaction as their versionable, and the table has no `versionDate`). This is what would catch a sixth delete being added later without a guard.

**`DatabaseCleanupSqlTest`** (vcell-rest `@QuarkusTest`) runs the real sweep against the testcontainers PostgreSQL with the real VCell schema from `scripts/init.sql`, and asserts behaviour rather than text: an orphan two hours old is collected, a row a save is still writing survives. Negative control — with the guard replaced by `1=1` it fails on exactly that second assertion (`a row a save is still writing must not be collected`). It also can't disturb other tests' fixtures, precisely because everything they create is under an hour old.

**Oracle** was verified by hand against `gvenzl/oracle-free:23-slim-faststart` (23ai):

```
SQL> SELECT SYSDATE AS now_dt, (SYSDATE - INTERVAL '1' HOUR) AS date1hr FROM DUAL;
NOW_DT               DATE1HR
2026-08-18 20:09:43  2026-08-18 19:09:43
```

plus all 15 generated Oracle statements executed against a stub schema (no `ORA-`, exit 0), and the same old/fresh orphan pair behaving as it does on PostgreSQL (`1 row deleted`, `save-in-flight` remaining).

I left the Oracle container **out of CI** rather than add a ~2 GB image pull to every run — happy to wire it in as a `regression.yml` group (merge queue + nightly) instead if you'd rather have it gated there. Your call.

jcschaff:
Oracle is now gated rather than spot-checked.

**`Oracle_IT` regression group** — `DatabaseCleanupOracleSqlTest` boots `gvenzl/oracle-free:23-slim-faststart` (23ai) on a stub schema and asserts three things: every statement `cleanupDatabase()` issues is accepted by Oracle, `SYSDATE - INTERVAL '1' HOUR` evaluates to exactly one hour back, and an orphan two hours old is collected while a row a save is still writing is spared. Negative control: with the guard replaced by `1=1` it fails on that last assertion, on Oracle as on PostgreSQL.

Verified end to end on this branch — `gh workflow run regression.yml --ref fix/cleanup-age-guard -f test-group=Oracle_IT`, [run 32183505458](https://github.com/virtualcell/vcell/actions/runs/32183505458):

```
✓ prepare regression matrix    3s
✓ CI-Test-group-Oracle_IT      2m35s   Tests run: 3, Failures: 0, Errors: 0
✓ regression-gate              4s
```

55s of that is the image pull plus Oracle boot on the runner — the cheapest group in the suite by some margin, and it costs the fast lane nothing. It runs in the merge queue before anything lands, nightly, and is selectable from the "Run workflow" dropdown.

Fast lane still green on the same commit: build, Fast-core, Fast-other, Quarkus, CodeQL.

Coverage now stands as: PostgreSQL on every push (real schema, real sweep), Oracle on the merge queue and nightly (real Oracle), and a database-free invariant test in `Fast` that fails if a sixth unreferenced-row delete is ever added without a guard.

jcschaff:
Filed #1994 for the broader version of this: the database layer branches on `DatabaseSyntax` in 55 places across 33 files, and CI only ever runs the PostgreSQL side while production only ever runs the Oracle side. `Oracle_IT` here closes that for the cleanup sweep; #1994 proposes closing it for the rest by running the existing Quarkus suite under Oracle as a second regression group.

Two findings from this PR that make it cheaper than it sounds: `SQLCreateAllTables.getVCellTables()` is public and `Table.getCreateSQL(DatabaseSyntax)` already emits per-dialect DDL, so no hand-written Oracle `init.sql` is needed; and `quarkus-jdbc-oracle`/`ojdbc11` are already vcell-rest dependencies. The blocker is `AgroalConnectionFactory.usePostgresql()` switching on the profile name — a new profile would hit its `default: throw`, so the datasource choice wants to be its own config property instead.

PR #2002, opened 2026-08-20
Using the langevin 1.4.13 solver

danv61:
Integrated the watchdog version of the langevin solver 1.4.13
Updating the slurm script for langevin batch execution
Fixing an even number of errors that were nullifying each other.
Commenting out (for now) the progress notifications sent directly from the script


jcschaff (approved):
looks good, Dan and I looked at it together.

PR #2042, opened 2026-09-01
Upgrade to Smart copy-paste dialog system

CodeByDrescher:
This PR is for a few different but related changes.
1) Smart copy/paste functionality via keybinds has been added to the Initial Conditions Panel.
2) In the math overrides panel, parameter scans now properly smart copy/paste.
3) The confirmation dialog for when smart copy/paste occurs has been improved visually and functionally.
4) Smart copy/paste actions now consistently use the Client Task Dispatcher
5) Logging improvements


danv61 (approved):
I am not sure about Nameable being so insular since most vcell entities implement getName and setName (anything that extends BioModelEntityObject or implements SymbolTableEntry or EditableSymbolTableEntry and more). 
Also, replacing equals with contains is unsafe in the code below:
if (evt.getPropertyName().equals(SelectionManager.PROPERTY_NAME_SELECTED_OBJECTS)) {
if (evt.getPropertyName().contains(SelectionManager.PROPERTY_NAME_SELECTED_OBJECTS)) {
I saw that you annotated the property with metadata, so I understand what you're doing, but it might have been more robust to do the contains first, and if true to really parse the property string, extract the exact property name (first token) and compare strict.
Nice feature!

PR #2049, opened 2026-09-02
Dan ss slurm2

danv61:
Watchdog fixed, was being called with the wrong arguments.
Concurrency cap relaxed to 30 simulations running concurrently (+1 watchdog)


jcschaff (approved):
reviewed together, approved.

PR #2080, opened 2026-09-10
Dan ss slurm3

danv61:
Relaxed max number of concurrent runs to 50 (51 including the watchdog)
Formal slurm attributes class (ServerInfo). Stub, using some predefined constants right now, we'll go to the server eventually and read there all the properties as required.
We need this to show multi-run info to the user in the Simulation Properties Panel: number of concurent runs, number of nodes used (at Les' request)
Number of concurrent runs is not user editable anymore.


vcell-core/src/main/java/cbit/vcell/solver/LangevinSimulationOptions.java
```diff
@@ -242,9 +242,9 @@ public void readVCML(CommentStringTokenizer tokens) throws DataAccessException {
 			} else if(token.equalsIgnoreCase(VCML.LangevinSimulationOptions_totalNumberOfJobs)) {
 				token = tokens.nextToken();
 				totalNumberOfJobs = Integer.parseInt(token);
-			} else if(token.equalsIgnoreCase(VCML.LangevinSimulationOptions_numberOfConcurrentJobs)) {
-				token = tokens.nextToken();
-				numberOfConcurrentJobs = Integer.parseInt(token);
+//			} else if(token.equalsIgnoreCase(VCML.LangevinSimulationOptions_numberOfConcurrentJobs)) {
+//				token = tokens.nextToken();
+//				numberOfConcurrentJobs = Integer.parseInt(token);
```
jcschaff, line 247: This is the blocking item. The chain below ends in `throw new DataAccessException("unexpected identifier " + token)`, so once this branch is gone every saved model that still carries the token fails to load. The token shipped in 8.1.6.02, so that is real user data, not just dev models.

Keeping the branch and discarding the value costs two lines and matches what `numOfParallelLocalRuns` already does six lines above.

```suggestion
			} else if(token.equalsIgnoreCase(VCML.LangevinSimulationOptions_numberOfConcurrentJobs)) {
				token = tokens.nextToken();		// not in use anymore, may be present in old VCML files; read and discard
```

Needs the `VCML` constant restored as well, suggested separately.

vcell-core/src/main/java/cbit/vcell/math/VCML.java
```diff
@@ -253,7 +253,7 @@ public class VCML {
 	public final static String LangevinSimulationOptions_Partition_Nx = "PartitionNx";
 	public final static String LangevinSimulationOptions_Partition_Ny = "PartitionNy";
 	public final static String LangevinSimulationOptions_Partition_Nz = "PartitionNz";
-	public final static String LangevinSimulationOptions_numberOfConcurrentJobs	= "NumberOfConcurrentJobs";
+//	public final static String LangevinSimulationOptions_numberOfConcurrentJobs	= "NumberOfConcurrentJobs";
```
jcschaff, line 256: Restoring this constant is half of the backward compatibility fix described in the review. The comment follows the pattern already used two lines below for `numOfParallelLocalRuns`.

```suggestion
	// deprecated, kept so old .vcml files containing this token still parse; value is read and discarded
	public final static String LangevinSimulationOptions_numberOfConcurrentJobs	= "NumberOfConcurrentJobs";
```

jcschaff (commented):
## Review

Read through the whole diff and checked the compatibility and client/server consistency questions against the repo. CI is green and it builds. One blocking item, one wrong number shown to the user, and a handful of smaller things.

### Blocking: saved Langevin models will fail to load

`LangevinSimulationOptions.readVCML` ends its token chain with a throw on anything unrecognized. This PR comments out the branch that consumed `NumberOfConcurrentJobs`, but that token has been written into every saved Langevin simulation since it was introduced, and it shipped:

| Fact | Value |
|---|---|
| Token added | 2025-09-30 |
| Shipped in | 8.1.6.02, tagged 2026-09-07 |

So any model saved by a released 8.1.x client now throws `DataAccessException: unexpected identifier NumberOfConcurrentJobs` on load. This is the persistence path, written at `SolverTaskDescription.getVCML` and read back at `readVCML`, not a debug dump.

The fix is already modelled three branches up: `numOfParallelLocalRuns` is kept as a read-and-discard branch with the comment "not in use anymore, may be present in some old VCML files". I have put one-click suggestions on both spots.

The XML side is fine. `XmlReader` uses `getChildText`, so it silently ignores an element it no longer reads.

### The node count shown to the user is off

Showing node count is the stated purpose here, so it is worth getting exact. The server divides the concurrent tasks *including* the watchdog, while the client divides the simulations only:

| Concurrent sims | Client displays | Slurm requests |
|---|---|---|
| 20 | 1 | 2 |
| 40 | 2 | 3 |
| 50 | 3 | 3 |

They agree only when the sim count is not an exact multiple of the per-node limit. Adding the watchdog task before dividing, the way `SlurmProxy` does, lines them up.

There is a second divergence behind that. The client reads the concurrency ceiling from a compiled-in constant, while `SlurmProxy` reads it from `PropertyLoader` and uses the constant only as a fallback. If the deployed property ever differs from the shipped client constant, the panel displays a concurrency and node count the server will not use. The `ServerInfo` stub is presumably where that gets fixed once it really consults the server, so this may just be worth a comment for now.

### Smaller points

- **`ServerInfo` was repurposed rather than extended.** The old contents were dead, I checked, so nothing breaks. But the name now sits in `cbit.vcell.server` while describing Langevin slurm tuning; something like `LangevinSlurmSettings` would say what it is. It also stays `Serializable` with no `serialVersionUID`, which will matter if it does start crossing the wire.
- **`nodes` is computed and then discarded** four times in `LangevinOptionsPanel`, at lines 92, 99, 160 and 769. Only `SimulationSummaryPanel` actually displays it.
- **The new `totalNumberOfJobs < 2` throw is unreachable.** `submitJob` only takes the batch path when that value already exceeds one. Fine as an assertion, it just can never fire.
- **Unused imports.** `LangevinLngvWriter` gained `PropertyLoader` and `ServerInfo` with no uses. The `SimulationSummaryPanel` import cleanup is correct, those really are gone.
- **Roughly sixty lines are commented out rather than deleted**, across six files, in a PR whose last commit is "Deleted unused functionality." Worth picking one convention.
- **Two identical copies** of `getRequestManager` and `getServerInfo`, about twenty five lines each, in the two panels.
- **Watchdog tick default halved** from 60 to 30 seconds. Worth flagging next to LangevinNoVis01 PR 44, where the alive message was commented out to cut spam. These two pull in opposite directions on message volume, so it may be worth deciding the intended rate across both.

Also worth knowing: the slurm fixture is unaffected by the 31 to 51 bump, since `TOTAL_JOBS=8` clamps below both. Nothing currently exercises the new ceiling.

https://claude.ai/code/session_01JJU2qv9M4ufLRy2em19yWK

vcell-core/src/main/java/cbit/vcell/solver/LangevinSimulationOptions.java
danv61: Good catch, I missed this one. I only checked that all is well with the xml reader / producer.

vcell-core/src/main/java/cbit/vcell/math/VCML.java
danv61: I deprecated them both.

danv61:
Fixed the vcml import regression, correct calculation of number of nodes, other smaller points fixed.
The spam is being cut by commenting out logger and messaging when nothing happen in the watchdog. Halfing the tick to 30 is intended.
I couldn't find a common ancestor class between LangevinOptionsPanel and SimulationSummaryPanel where to place single instances of getRequestManager() and getServerInfo(), hence the duplicated code

jcschaff (approved):
Looks good Dan, approved - my comments were addressed.

